PR #130, opened 2020-04-25
Turn COUNTRY_THAILAND into a bool

igfoo:
For now at least, the value is set based on the `COUNTRY_THAILAND` preprocessor symbol, so the compiler will optimise out all the code that isn't being used, and performance will be unchanged.

But it would now be easy to make it a run-time parameter instead.

Doing it this way also means that the compiler can see all the code, so it is less likely that code changes for UK/US/... will break the Thailand code.

Review will probably be easier if you [ignore whitespace differences](https://github.com/mrc-ide/covid-sim/pull/130/files?w=1) as some code needed to be indented deeper.


igfoo:
Closing in favour of #141

PR #139, opened 2020-04-27
python codes cleanup

jnozsc:
tiny style improvement


matt-gretton-dann:
Thank you for this contribution.  Apologies for the delay in responding, but we are still extremely busy and responses will be delayed.

I believe #159 mostly covers most of these changes.  Although I will `flake8` that change to make sure.  So closing.

PR #149, opened 2020-05-01
Dlaydon/more documentation to code

dlaydon:
Hey @matt-gretton-dann, I added/deleted a few comments and renamed a couple of variables that would be good to merge before everything diverges too much. Cheers, Danny


igfoo:
Looks good to me!

PR #152, opened 2020-05-02
Adding Binder Badge

kmader:
Adding a binder badge to make it easier to run the code without installing anything using mybinder.org
- added r and tidyverse for running rscripts
- added matplotlib and jupyter

The badge doesn't work until the code is merged but you can test the code at https://mybinder.org/v2/gh/kmader/covid-sim/patch-1

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# COVID-19 CovidSim Model
+# COVID-19 CovidSim Model [![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/mrc-ide/covid-sim/master)
 
 This is the COVID-19 CovidSim microsimulation model developed by the MRC Centre for Global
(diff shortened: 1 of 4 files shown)


matt-gretton-dann:
Thank you for the contribution.  We are currently still very busy and so responses will be delayed.

I am hesitant to accept this as my reading of the resources available to the Binder docker containers is that they only have 2GB of memory, which I believe means even the regression tests can not run to completion.  Therefore, it will not be very useful for most use cases.  What scenarios have you managed to run?

kmader:
I hadn't realized the 2GB limitation, maybe it makes more sense to have as a colab or kaggle kernel. I was just trying to make it easier to run without setting everything up locally.

matt-gretton-dann:
Closing PR following discussion.

Issue #75, opened 2020-04-09
Integrate regression tests into cmake


matt-gretton-dann:
The regression tests are currently integrated into CI in an ad hoc manner.  We should add them as testing targets in CMakeLists.txt to enable a `make test` target and improve issues.  This will make testing scalable and easy to run.

It is dependent on #37 so we only have one binary.

matt-gretton-dann:
Done

PR #153, opened 2020-05-04
Ci love

cjgdev:
This pull request integrates the regression tests with ctest, as referenced in ticket #75 

Regression tests are run via ctest, and can be disabled by passing -DENABLE_REGRESSION_TESTS:BOOL=OFF when configuring cmake, the default is enabled.

This changeset also includes a dockerignore file, which makes building the Docker images a good bit faster.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,5 @@
+.git
+.github
+.vscode
+out
+**/build*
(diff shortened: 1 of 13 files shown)


matt-gretton-dann:
Closing this in favour of #159 which merges the test suite drivers into a single piece of code.

PR #154, opened 2020-05-04
Added support for clang tooling [analyze,format,tidy,sanitizers]

cjgdev:
This pull request adds support for code linting, static analysis, format checking and runtime sanitizers, provided by clang.

Since this is only the initial introduction, this pull request does not fix any code issues raised by the tools, since this would be a lot of work and require making a potentially large number of edits to existing (working) code.


cjgdev:
Hello Matthew, great to hear from you, hello and thank you for taking the time to review my pull request. There's a bit to unpack in your feedback, but I'll do my best to cover everything.

- Yes I have some documentation prepared already, I'll merge that into this changeset.

- Toolsets are a feature of the CMake program, so the top-level CMakeLists.txt doesn't need any modification.

- I was very much headed in the direction of integrating these tools into the CI system, and this was the first step, so because it sounds like there is notional interest I shall continue down that route on this branch.

- I understand your point, but integrating nicely with Visual Studio under Windows is a *bit* of a tricky beast (read: let's not go there), but as you suggested yourself, the CI checks are a natural fit for where these might slot into your dev / build process. The types of checks these tools perform should apply equally well to any platform / compiler, so there is value in having that.

weshinsley:
Very late to the party on this - but see (Issue #372), -fsanitize=thread is a false-positive-fest on Linux unless you have already compiled OpenMP with the necessary hooks (in which case you get all-clear).

cjgdev:
Hi Wes, thanks for coming back to this pull request, I had all but forgotten about it.

I am afraid that the window of opportunity for me to participate in any meaningful way has unfortunately been closed due to furlough ending a fortnight ago. I just wish I could have done more to help, but that simply wasn’t possible.

I will close the pull request. Manual method is good enough for covid-sim.

weshinsley:
Hi @cjgdev - no worries - sorry it took so long for me to get to this, and many thanks for your thoughts earlier; we will try to pick them all up eventually one way or another.

PR #164, opened 2020-05-06
Various documentation fixes

ozmorph:
Hello all,

I've attempted to document many of the missing command-line arguments for the covidsim binary in the inputs-and-outputs.md file. I've also tried to more accurately identify which parameters are required and which are optional. Argument examples now point to the files in the data folder, when applicable, so that they have more meaning other than a random string name.

I've also added a new glossary.md file to hyperlink medical and scientific terms found in other doc files. This is to help an average C++ developer (like myself) that may not immediately understand the terminology used in the code but are looking to contribute to the project.

In addition, I've made some fixes to the various Markdown files throughout the project that were discovered by the [markdownlint](https://github.com/markdownlint/markdownlint) tool including:

1. <> brackets for links instead of relying on GitHub to automatically detect and render URLs.
2. using a standard heading hierarchy starting at a single `#`
3. using the dash (-) character for unordered lists (it was mix between *, -, and + before)
4. adding the language to most fenced code blocks to get syntax highlighting where applicable
5. adding blank lines where generally applicable (between paragraphs and lists, after headings, etc.)

There were a few minor spelling mistakes corrected (none related to -ize/-ise spellings, I assure).

Finally, I've done my best to reformat most of the Markdown files to use a standard column width of 80 characters so that it is easier to edit/grep through the documentation. I tried my best to break lines on natural boundaries within or at the end of sentences. This would also allow minor updates to documentation to be better tracked/grepped through Git in the future.

Let me know if there are any changes/omissions you would like me to make to this PR!


ozmorph:
Rebased with upstream master.

PR #200, opened 2020-05-09
Changed Cells, CellLookup, MCells, and MCellsLookup to std::vector(...)

chrisatang:
Changed the Cells, CellLookup, MCells, and MCellsLookup from dynamic calloc(...) to std::vector(...), which is a start to resolving any possible memory issues the current code might have.

### DIFF
--- a/src/CovidSim.cpp
+++ b/src/CovidSim.cpp
@@ -75,7 +75,7 @@ std::vector<cell> Cells(P.NC); // Cells.at(i) is the i'th cell
 std::vector<cell*> CellLookup(P.NCP); // CellLookup[i] is a pointer to the i'th populated cell
 std::vector<microcell> Mcells(P.NMC);
 std::vector<microcell*> McellLookup(P.NMCP);
-std::vector<std::vector<place>> Places;
+place** Places;
 adminunit AdUnits[MAX_ADUNITS];
 //// Time Series defs:
 //// TimeSeries is an array of type results, used to store (unsurprisingly) a time series of every quantity in results. Mostly used in RecordSample.
--- a/src/CovidSim.cpp
+++ b/src/CovidSim.cpp
@@ -71,10 +71,11 @@ param P;
 person* Hosts;
 household* Households;
 popvar State, StateT[MAX_NUM_THREADS];
-cell* Cells; // Cells[i] is the i'th cell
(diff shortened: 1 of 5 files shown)


src/CovidSim.cpp
```diff
@@ -71,9 +71,10 @@ param P;
 person* Hosts;
 household* Households;
 popvar State, StateT[MAX_NUM_THREADS];
-cell* Cells; // Cells[i] is the i'th cell
-cell ** CellLookup; // CellLookup[i] is a pointer to the i'th populated cell
-microcell* Mcells, ** McellLookup;
+std::vector<cell> Cells(P.NC); // Cells.at(i) is the i'th cell
+std::vector<cell*> CellLookup(P.NCP); // CellLookup[i] is a pointer to the i'th populated cell
+std::vector<microcell> Mcells(P.NMC);
+std::vector<microcell*> McellLookup(P.NMCP);
```
matt-gretton-dann, line 77: You are trying to initialise these at program start up time when `P.NC` is `0` - so this will fail - I think need to have something like:

```c++
std::vector<cell> Cells;

// Whereever the calloc was
Cells.resize(P.NC);
```

etc.

src/CovidSim.cpp
```diff
@@ -2567,25 +2568,25 @@ void InitModel(int run) // passing run number so we can save run number in the i
 	{
 		for (i = tn; i < P.NC; i+=P.NumThreads)
 		{
-			if ((Cells[i].tot_treat != 0) || (Cells[i].tot_vacc != 0) || (Cells[i].S != Cells[i].n) || (Cells[i].D > 0) || (Cells[i].R > 0))
+			if ((Cells.at(i).tot_treat != 0) || (Cells.at(i).tot_vacc != 0) || (Cells.at(i).S != Cells.at(i).n) || (Cells.at(i).D > 0) || (Cells.at(i).R > 0))
```
matt-gretton-dann, line 2571: `std::vector::at()` will throw an exception on a bad access. Not a bad thing per se - however, the code isn't exception aware and I think there will be problems with exceptions crossing thread boundaries.

There are two solutions to this:

The easy one: Just use `[]` access everywhere.  (Why do you use it in some places and not others?).

The harder one, put `try`/`catch` blocks everywhere that is appropriate.

PR #204, opened 2020-05-09
Drop #pragma once

zdroid:
Some of the heades contain both `#pragma once` and `#ifndef` guards. Since all headers already use `#ifndef`, I guess this solution is easier.


Feynstein:
I think that the opposite would be better because it's cleaner, and since there aren't that much headers it should be easy to do.

zdroid:
I was thinking about that, but didn't want to interfere into the opinions of the developers here. :smile:
It can be changed that way too, if they want it.

PR #207, opened 2020-05-10
Ensure consistency of the C++ standard by setting it to C++ 14

zebmason:
Visual Studio 2019 has minimum standard of C++14, so set this explicitly
gcc does not fully support the C++17 standard but has supported C++14 for some time
Currently CMakeLists.txt specifies C++11 leading to potential issues from people working on Windows.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,7 +45,7 @@ else()
 endif()
 
 # specify the C++ standard
-set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD 14)
 
 add_subdirectory(src)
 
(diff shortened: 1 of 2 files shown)


cjgdev:
> gcc does not fully support the C++17 standard

Can you please share the platform / compiler / version where `-std=c++17` didn't work? You will see from the [C++ Standards Status](https://gcc.gnu.org/projects/cxx-status.html) that C++17 support has been there since GCC 7, and the project is now 3 major releases along from that. I would have thought C++17 would have been the common baseline between compilers this recent.

zebmason:
@cjgdev Christopher the [GCC support page](https://gcc.gnu.org/projects/cxx-status.html#cxx17) states "Some library features are missing or incomplete"

zebmason:
@matt-gretton-dann Matthew could you merge this PR, please?

matt-gretton-dann:
@cjdev: Ubuntu 18.04 (which is still I believe the ubuntu-latest image on GitHub Actions) ships with GCC-7 which doesn't support -std=c++17 (it does support -std=c++1z) and has some missing functionality.

PR #213, opened 2020-05-11
Refactor out point classes, etc.

zebmason:
Create Vector and BoundingBox classes to simplify code.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 # CMakeLists.txt
 
 # CMake setup
-cmake_minimum_required (VERSION 3.8)
+cmake_minimum_required (VERSION 3.11)
 
 # Project initialisation
 project("CovidSim")
@@ -47,8 +47,13 @@ endif()
 # Specify the C++ standard
 set(CMAKE_CXX_STANDARD 14)
 
+include_directories(${CMAKE_CURRENT_SOURCE_DIR}/include)
+
+add_subdirectory(third_party)
+add_subdirectory(include)
 add_subdirectory(src)
 
+add_subdirectory(unit_tests)
 if(WIN32 AND USE_WIN32_BMP)
   message(WARNING "USE_WIN32_BMP is enabled.  Therefore testing is disabled.")
   message(WARNING "To enable testing invoke with `cmake -DUSE_WIN32_BMP=NO`")
--- a/include/CMakeLists.txt
+++ b/include/CMakeLists.txt
@@ -1,0 +1,1 @@
+add_subdirectory(geometry)
--- a/include/geometry/CMakeLists.txt
+++ b/include/geometry/CMakeLists.txt
@@ -1,0 +1,2 @@
+# Set up the IDE
+set(GEOMETRY_HDR_FILES Geometry.h)
--- a/include/geometry/Geometry.h
+++ b/include/geometry/Geometry.h
@@ -1,0 +1,272 @@
+#pragma once
+
+#include <Eigen/Dense>
+
+namespace CovidSim
+{
+  namespace Geometry
+  {
+    typedef Eigen::Vector2i Vector2i;
+    typedef Eigen::Vector2d Vector2d;
+    typedef Eigen::Vector2f Vector2f;
+
+    typedef Eigen::DiagonalMatrix<double, 2> DiagonalMatrix2d;
+
+    /// General class for representing a 2D maximum point
+    template<typename T>
+    class Maximum2 : public Eigen::Matrix<T, 2, 1>
+    {
+    public:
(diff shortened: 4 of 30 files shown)


matt-gretton-dann:
Thank you very much for your PR.  Please be aware that we are very busy and so reviews may take a while - especially as this is a re-engineering PR.  Also please be aware that the model will be developing underneath you and so you may have to rebase/merge master into your PR on a semi-regular basis to keep the code up to date.

Having said that :smile:,  I generally like the direction of this PR.

As this is the first PR that will use C++ classes there is plenty of bike-shedding that can be done, but I will try to keep that to a minimum - but let's try to document some coding styles as we go along to keep consistency with future changes.

So my initial comments for discussion - this is not a complete or detailed review and I'm sure I will have some more comments as we go along:

 1. Whilst I like the `namespace`s - I'm not a fan of one header per class, I think Vector.h and BoundingBox.h could be one header 'Geometry.h'.
 2. Your classes feel like a cross between a POD and a 'full' class, and it feels wrong - I think we should either:
  - "Full class": Make all member variables private, and use constructors, getters/setters, and other member variables.
  - "POD": Keep the member variables public, and the member methods you've created should be non-members.  So the `structs` are just data.
 3. Why `BoundingBoxD` as a derived class from `BoundingBox<T>`?  Why not just have all the methods in `BoundingBox<T>`?
 4. Doxygen style comments on every class, variable, and method/function please.
 
Bike-shedding on naming conventions:
 * Classes & Namespaces: CamelCase
 * Member variables: snake_case_with_trailing_underscore_
 * Member methods: snake_case()
 * Variables: snake_case

Once again thank you for your effort and time.

zebmason:
I'll make some changes tomorrow. The idea was to temporarily replace the existing code with code that is easier to replace, e.g. a third party library for GIS work however I would be wary of reviewing the candidates myself. Obviously temporary solutions can become rather permanent.

BoundingBoxD is derived because it contains actual types, e.g. adding 1.e-6 implies T is double.  You'll note the comment on BoundingBoxD::Outside suggests there is an obscure bug on equality. So hack code consistent with the design intent which is to replace existing code with code that doesn't fail the regression tests.

W.R.T. naming conventions is a [ClangFormat](https://clang.llvm.org/docs/ClangFormat.html) file fine? Do you want to push one or should I?

matt-gretton-dann:
> I'll make some changes tomorrow. The idea was to temporarily replace the existing code with code that is easier to replace, e.g. a third party library for GIS work however I would be wary of reviewing the candidates myself. Obviously temporary solutions can become rather permanent.

OK.

> BoundingBoxD is derived because it contains actual types, e.g. adding 1.e-6 implies T is double. You'll note the comment on BoundingBoxD::Outside suggests there is an obscure bug on equality. So hack code consistent with the design intent which is to replace existing code with code that doesn't fail the regression tests.

OK - that makes sense.

> W.R.T. naming conventions is a [ClangFormat](https://clang.llvm.org/docs/ClangFormat.html) file fine? Do you want to push one or should I?

Please leave that to me.  Its been on our to-do list for a while, but there are still significant model developments happening and so it has never been quite the right time to do a "reformat the whole source" PR.

zebmason:
@matt-gretton-dann Build is failing because I have unit tests and need to do a git submodule update before the build starts. I see that you look after the Dockerfile so seem to be my best bet.

zebmason:
Might actually be my fault

zebmason:
Have commented out the CMake descent into the unittests directory and all the Linux builds pass so I presume the Windows one will when it eventually finishes. My error earlier was forgetting to commit the submodule directory but that made no difference...

matt-gretton-dann:
Can you do the following trick with CMake to get it to install GoogleTest at build time?

```cmake
include("FetchContent")
FetchContent_Declare(
  googletest
  GIT_REPOSITORY https://github.com/google/googletest.git
  GIT_TAG        release-1.10.0
)

FetchContent_GetProperties(googletest)
if(NOT googletest_POPULATED)
  FetchContent_Populate(googletest)
  set(gtest_force_shared_crt ON CACHE BOOL "" FORCE) 
  add_subdirectory(${googletest_SOURCE_DIR} ${googletest_BINARY_DIR})
endif()
include(GoogleTest)
```

Although I think that will move our minimum version requirement to 3.11 which is not available OOB on Ubuntu 18.04 (which is on 3.10).  The [GoogleTest README](https://github.com/google/googletest/blob/master/googletest/README.md) has some descriptions on how to do this integration without requiring submodules.

zebmason:
Most of this has been pulled over as other PRs. I think the main thing remaining is the use of Eigen3 for delayed evaluation of vector algebra.

PR #215, opened 2020-05-11
Drop SharedFuncs.h

zdroid:
Having an additional header file for one function declaration is probably not worth the cost. `SharedFuncs.h` included only `void UpdateProbs(int)` which can be put in a more common header, `SetupModel.h`.


Bmirt:
Bad Practice, Rejected

zdroid:
@matt-gretton-dann done

PR #216, opened 2020-05-11
Update gitignore file for some IDEs

YsHaNgM:


.gitignore
```diff
@@ -15,3 +15,31 @@ build_*/
...
+/.cproject
+/.project
+# QtCreator config files:
+/*.config
+/*.creator
+/*.creator.user
+/*.files
+/*.includes
+/*.autosave
+/CMakeLists.txt.user*
```
matt-gretton-dann, line 36: We don't use any of these so I don't see a need for them in the .gitignore.

.gitignore
```diff
@@ -15,3 +15,31 @@ build_*/
...
+# Mac Finder:
+.DS_Store
+# CLion/IntelliJ
+*cmake-build-debug
+.idea/*
```
matt-gretton-dann, line 41: Ditto

.gitignore
```diff
@@ -15,3 +15,31 @@ build_*/
...
+# VScode
+.vscode/
+.vscode/*
```
matt-gretton-dann, line 44: Isn't `.vscode/` enough here?

.gitignore
```diff
@@ -15,3 +15,31 @@ build_*/
...
+build/
```
matt-gretton-dann, line 45: This is an issue earlier in the .gitignore file - the
```
build
```
line further above should be
```
build/
```

.gitignore
```diff
@@ -15,3 +15,31 @@ build_*/
 .project
 *.vcxproj.user
 tests/*-output/
+
+
+*.pyc
+*~
```
matt-gretton-dann, line 21: Duplicate.

.gitignore
```diff
@@ -15,3 +15,31 @@ build_*/
 .project
 *.vcxproj.user
 tests/*-output/
+
+
```
matt-gretton-dann, line 19: Only one line gap

zdroid:
If gitignore should be updated, then it should be updated according to `https://github.com/github/gitignore/blob/master/C%2B%2B.gitignore`. Editor or IDE specific files should belong to your own system's gitignore.

.gitignore
```diff
@@ -9,37 +9,17 @@ x64
 *.bin
 wpop_*.txt
 out/
-build
```
zdroid, line 12: This is a useful directory...

YsHaNgM: I suppose this was a issue mentioned above? Can you explain more?

zdroid: It isn't. Official `build.md` doc also suggest using `build/`.

YsHaNgM: That indicates to create a new build directory. But not include it in source code repository, neither CMakeLists added build subdirectory.

zdroid: Not really: https://github.com/mrc-ide/covid-sim/blob/master/docs/build.md#building-with-makefiles

YsHaNgM: Yes you should have build directory on your own behalf which is a CMake convention.

zdroid: When you clone the repository, cd to it and execute the command, you'll have `build/` directory inside the git repository. This PR would make a lot of unwanted files to be tracked by Git. I still don't understand the need to remove this line.

matt-gretton-dann: I think what is being missed is the line should be:

```suggestion
build/
```

That is with a trailing `/` to indicate it is a directory.

YsHaNgM:
That’s why added `build/` underneath.

.gitignore
YsHaNgM: I see. I had this change locate under vscode section. I have changed it back now. Hope it would not mislead to the disscuss above this time.

zdroid:
@YsHaNgM my apologies then, really missed that one :)

PR #217, opened 2020-05-11
Rename binio to BinIO

zdroid:
To be consistent with other file names. Also removes a mention of non-existent `src/avi_utils.h`.


zdroid:
@matt-gretton-dann Done.

PR #222, opened 2020-05-11
Add kernel type enum

endoffile78:


src/Kernels.h
```diff
@@ -1,6 +1,18 @@
 #ifndef COVIDSIM_KERNELS_H_INCLUDED_
 #define COVIDSIM_KERNELS_H_INCLUDED_
 
+enum KernelType {
```
cjgdev, line 4:
```suggestion
enum class KernelType {
```
Since this is a C++11 codebase, the use of enum classes should be preferred. Casts to / from int will need to be made explicit, but ultimately leads to fewer surprises in the long-run.

endoffile78: Done

src/Kernels.h
```diff
@@ -1,6 +1,16 @@
...
 
+enum class Kernel_Type {
+    KERNEL_EXPONENTIAL = 1,
+    KERNEL_POWER = 2,
+    KERNEL_GUSSIAN = 3,
+    KERNEL_STEP = 4,
+    KERNEL_POWERB = 5, // figure out a better name, power b isn't descriptive
+    KERNEL_POWERUS = 6,
+    KERNEL_EXPONENTIAL_POWER = 7,
+};
```
matt-gretton-dann, line 12: If using `enum class` no need to put `KERNEL_` as a prefix to each value as we have to use the class name to qualify the value.

src/Kernels.h
```diff
@@ -1,6 +1,16 @@
 #ifndef COVIDSIM_KERNELS_H_INCLUDED_
 #define COVIDSIM_KERNELS_H_INCLUDED_
 
+enum class Kernel_Type {
```
matt-gretton-dann, line 4: `KernelType` please (no `_`).

src/Kernels.cpp
```diff
@@ -28,22 +28,32 @@ void InitKernel(int DoPlaces, double norm)
...
+        case Kernel_Type::KERNEL_POWERUS:
+            Kernel = PowerKernelUS;
+            break;
+        case Kernel_Type::KERNEL_EXPONENTIAL_POWER:
+            Kernel = PowerExpKernel;
+            break;
+        default:
+            ERR_CRITICAL_FMT("Unknown kernel type %d.\n", P.KernelType);
+    }
+
```
matt-gretton-dann, line 56: If we're going to move to a `switch` table would a lookup table be better?

So something like:

```c++
if (in_range(P.KernelType)) {
  Kernel = KernelFn[P.KernelType];
}
else {
  ERR_CRITICAL_FMT("Unknown kernel type %d.\n", P.KernelType);
}
```

If doing that then maybe rebase the enum from 0 - although that needs checking its safe - I haven't done that.

matt-gretton-dann:
This has been replaced with #307 .

PR #226, opened 2020-05-12
Remove redundant double to int conversions

zdroid:
In `Dist.cpp` `xi` and `yi` are declared as double and converted to int 10 times, while being subtracted only 4 times from doubles. Instead of that, `xi` and `yi` can be outright declared as `int`. Based on my understanding, this should be more efficient.

### DIFF
--- a/src/Dist.cpp
+++ b/src/Dist.cpp
@@ -7,13 +7,14 @@
 
 double sinx[361], cosx[361], asin2sqx[1001];
 
-//// **** //// **** //// **** //// **** //// **** //// **** //// **** //// **** //// **** //// **** //// **** //// **** //// **** //// **** //// **** //// **** //// **** //// **** 
+//// **** //// **** //// **** //// **** //// **** //// **** //// **** //// **** //// **** //// **** //// **** //// **** //// **** //// **** //// **** //// **** //// **** //// ****
 //// **** DISTANCE FUNCTIONS (return distance-squared, which is input for every Kernel function)
-//// **** //// **** //// **** //// **** //// **** //// **** //// **** //// **** //// **** //// **** //// **** //// **** //// **** //// **** //// **** //// **** //// **** //// **** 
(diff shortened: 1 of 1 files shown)


cjgdev:
Have you benchmarked this version against the previous? If so, then how does this version compare?

It might be interesting to evaluate whether activating enhanced instructions and tuning for the target processor would achieve the same or better results, in which case this might be better delegated to the compiler optimiser.

For gcc `-O4` is activated for the Release configuration, and it might be useful to compare against `-O4 -march='<arch>' -mtune='<arch>'` (where `<arch>` corresponds to your particular CPU architecture, as identified in the [gcc x86-options documentation](https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html)).

zdroid:
I wanted to, but the function uses functions and constants that are defined in various other files. A proper run would probably require me to run the whole simulation, which requires too much RAM. Is there a way to infer the performance from tests, for example?

This function surely can be further optimized to use less declarations and less conversions. For example, `y = y * y` line can be removed and replaced in `x = fabs(1000 * (y + x * cy1 * cy2));` as `x = fabs(1000 * (y * y + x * cy1 * cy2));`. Do you have an estimate on how many times this function is used in the simulation? It seems like simplifying even one operation would give a significant efficiency boost.

zdroid:
In this PR, double to int conversion is replaced with int to double (which is supposed to be far safer).

Abstracting sounds like a good idea, as it'd simplify the code. You can propose that change first, then I can see if some change like the one I proposed in this PR still applies.

matt-gretton-dann:
> In this PR, double to int conversion is replaced with int to double (which is supposed to be far safer).

This PR is doing the `double` to `int` conversion by hand ourselves.  And therefore I have to convince myself that we aren't losing anything by doing this conversion (I don't think we are but I have to spend some time to convince myself0).

zdroid:
Closing this until I find a better solution.

PR #227, opened 2020-05-12
Constexpr for PI and EARTHRADIUS plus Documentation

ozmorph:
I'm a strong proponent of using constexpr in scientific computing where possible as it encourages less reliance on so-called "magic" numbers and instead places emphasis on mathematically provable values or values that have been standardized (such as in C++20's std::numbers).

I documented where the magic `6366707.0` value of EARTHRADIUS came from and rewrote it to derive the same result through simple math equations. Using a calculator, the value that should be generated at compile time is `6,366,707.019675681`.

This pull request does not require an update to the hashes of the regression tests. However if precision of the PI constant is extended to the full 64-bit precision, then I imagine this could affect the results of the simulation.

### DIFF
--- a/src/Constants.h
+++ b/src/Constants.h
@@ -1,8 +1,67 @@
 #ifndef COVIDSIM_CONSTANTS_H_INCLUDED_
 #define COVIDSIM_CONSTANTS_H_INCLUDED_
 
-#define PI 3.1415926535
-#define EARTHRADIUS 6366707.0
+/**
+ * Math constant defined as the ratio of a circle's circumference to its diameter.
+ *
+ * TODO: since all calculations using this constant are being automatically
+ * type-casted to double, should the precision be extended for more accuracy in
+ * the simulations?
+ *
+ * Eventually could be replaced with C++20's std::numbers::pi.
+ * https://en.cppreference.com/w/cpp/header/numbers
+ */
+constexpr double PI = 3.1415926535; // full double precision: 3.14159265358979323846
+
+/**
+ * An arc minute of latitude along any line of longitude in meters.
+ *
+ * Also known as the International Nautical Mile.
+ *
+ * @see https://en.wikipedia.org/wiki/Nautical_mile
+ */
+constexpr int NMI = 1852;
+
+/**
+ * The number of arc minutes in one degree.
+ *
+ * @see https://en.wikipedia.org/wiki/Minute_and_second_of_arc
+ */
+constexpr int ARCMINUTES_PER_DEGREE = 60;
+
+/**
+ * The number of degrees in a complete rotation.
+ *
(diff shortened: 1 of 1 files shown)


zdroid:
I support dropping defines and using const/constexpr instead. But is it actually required to do all this? Earth's radius is not really a magical constant. It's a well defined and precisely calculated concept in astronomy.

ozmorph:
@zdroid Thanks for the feedback, I do agree with your sentiment. The granularity level at which I went to is certainly not required at this current point in time, and I would be open to a better solution if there is one.

Playing devil's advocate here and to explain my original intuition, the original constant `6366707` does not show up on in any official astronomical standard as the size of the Earth in meters. It does however show up in a book titled `Mechanics of Flight` which uses a spherical approximation, which is what I documented in this PR. [Wikipedia](https://en.wikipedia.org/wiki/Earth_radius) says `6378100` meters for SI. [NASA's Goddard](https://nssdc.gsfc.nasa.gov/planetary/factsheet/earthfact.html) and [IUGG](https://en.wikipedia.org/wiki/International_Union_of_Geodesy_and_Geophysics) say `6378137` meters. The original magic value is within the range of generally accepted values of the radius, but it is statistically different (1%) to what has been precisely calculated. Somewhat unrelated but still noteworthy in the case of deterministic simulations (such as this repository with RNG seeds), choosing to use non-precise initial conditions/constants could lead to large [butterfly effects](https://en.wikipedia.org/wiki/Butterfly_effect) that affect the results.

Therefore I argue that allowing mathematical constants to be derived from their most basic values can remove the artificially limited precision that magical constants sometimes incur on scientific simulations. We are using doubles after all, so why not allow the extra precision to be used when the underlying platform is capable of it?

My only argument for defining `ARCMINUTES_PER_DEGREE` and `DEGREES_PER_TURN` would eventually be to see other code in this repo that uses calculated random projections in Cartesian coordinates to be precise and use the same sound and documented reasoning rather than depending on magical (perhaps inaccurate) constants such as [<b>this</b>](https://github.com/mrc-ide/covid-sim/blob/6abcaa3b51a984932e150547545b2c4c15f0a5c4/src/SetupModel.cpp#L1215) one in SetupModel.cpp.

However if this is obtuse and unnecessary to the majority of people, then I will remove the derivations and simplify my PR down to a well-documented constexpr value that explains where the value comes from.

zdroid:
@ozmorph I had this one in mind: R_NeE = 6.3781 × 10^6 m according to [IAU](https://arxiv.org/pdf/1510.07674.pdf). Your point is valid, that even a minimal change can lead to a butterfly effect. Authors of this project probably can help more with this issue. I guess it doesn't change the evolution of the model much, since significant interactions are mostly short-range.

PR #232, opened 2020-05-12
Remove zfwrite/read_big

zdroid:
Since the functions are equivalent to `fwrite/read_big`, they can be dropped.


src/binio.cpp
```diff
@@ -14,7 +14,7 @@ size_t fwrite_big(void *buffer,size_t size,size_t count, FILE *stream)
 
 	st=mx/((unsigned long long) size);
 	n=count/st;
-	for(j=0;j<n;j++) 
+	for(j=0;j<n;j++)
```
matt-gretton-dann, line 17: Please drop these stray white-space changes (here and below).

zdroid: Aren't those white spaces extraneous? My editor automatically strips them (on save). Many editors will do the same. But if you want, I can remove those changes.

matt-gretton-dann: Please remove these changes.  They are unrelated to the purpose of the commit and so will cause confusion when looking at the log in the long-term.

matt-gretton-dann:
Please also rebase to fix conflicts.

zdroid:
Okay, should be done now.

zdroid:
I dropped a define that's not needed anymore.

PR #234, opened 2020-05-12
Avoid ternary operator in StepKernel

zdroid:
The statement is logically equivalent to the one I proposed. `<=` probably can be replaced with `<`, since it's the comparision of two doubles.

### DIFF
--- a/src/Kernels.cpp
+++ b/src/Kernels.cpp
@@ -100,7 +100,7 @@ double GaussianKernel(double r2)
 }
 double StepKernel(double r2)
 {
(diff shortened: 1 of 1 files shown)


zdroid:
Does it really? I've never seen ternary operation used to return two bools. It's more confusing that way to be honest.

matt-gretton-dann:
But its not returning a `bool`.  The function returns a `double` - which is why being explicit about the values is better in my opinion.  I don't have to think about what conversions are taking place, and I'm explicitly showing that its not the "truthiness" I am interested in but the values 1 and 0 for the step function.

PR #236, opened 2020-05-12
Removed unused variables.

CloneDeath:
Removed some unused variables, either they were only ever assigned and never read, or they were never assigned to begin with.

I also added the .idea directory to the .gitignore (used by JetBrains CLion IDE, only holds temporary user workspace data, like which files the user currently has open).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ build_*/
 .project
 *.vcxproj.user
 tests/*-output/
+.idea/
 
 # VScode
 .vscode/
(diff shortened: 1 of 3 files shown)


CloneDeath:
@matt-gretton-dann No problem, done. :smile:

CloneDeath:
`Could NOT find Python3 (missing: Python3_EXECUTABLE Interpreter)`

Looks like the failure is because python3 isn't found on one of the test machines, not related to any change I have made.

I'm seeing this same failure in a couple other PR's. Unfortunately, I don't think there's anything I can do...

PR #237, opened 2020-05-13
Set up the source and header files for the IDE

zebmason:
Set the source_group for the IDE

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,7 +1,12 @@
 # CMakeLists.txt for src directory
 
+# Set up the IDE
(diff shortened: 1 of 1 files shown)


zdroid:
There is also https://stackoverflow.com/questions/3201154/automatically-add-all-files-in-a-folder-to-a-target-using-cmake, so that may be the better option

zebmason:
@zdroid Bemused as my commit sets the folder for the IDE a la vcxproj.filters file, it doesn't create the list of files which is what is discussed on stackoverflow

zdroid:
Your proposal still manually lists all `.cpp` and `.h` files, while the one on the link above lists them all automatically. What I wanted to say is that this project may use that approach to further simplify what you proposed. :wink:

zebmason:
@matt-gretton-dann Agree as I've been caught out in the past by doing a 
file(GLOB ...

PR #244, opened 2020-05-13
Fix kernel index issues with large geographies (Russia in particular)

matt-gretton-dann:
We are getting `r too large in NumKernel` errors when processing the Russia geography.  This PR fixes the issue.

There are two issues:

 1. Russia spans -180/180 degrees longitude and so the cut we do to flatten the earth causes it to have the wrong geometry in the model.  Add `[Longitude cut line]` parameter to enable us to move where we cut the earth to give Russia a contiguous geometry.

 2. We were calculating the longest distance possible in the bounding box of the geographies as the distance bottom-left to top-right.  This may not hold on a sphere.  Check whether Going along or up the sides of the bounding box will give us a longer distance, and use that.

@weshinsley : I have only updated Russia's admin file to handle the -180/180 cut - you may want to review the population density files to see if there are any other countries whose admin files need a `[Longitude cut line]` parameter.

### DIFF
--- a/src/SetupModel.cpp
+++ b/src/SetupModel.cpp
@@ -152,6 +152,12 @@ void SetupModel(char* DensityFile, char* NetworkFile, char* SchoolFile, char* Re
 		fprintf(stderr, "Adjusted bounding box = (%lg, %lg)- (%lg, %lg)\n", P.SpatialBoundingBox[0], P.SpatialBoundingBox[1], P.SpatialBoundingBox[2], P.SpatialBoundingBox[3]);
 		fprintf(stderr, "Number of cells = %i (%i x %i)\n", P.NC, P.ncw, P.nch);
 		fprintf(stderr, "Population size = %i \n", P.PopSize);
+		if (P.width > 180) {
+			fprintf(stderr, "WARNING: Width of bounding box > 180 degrees.  Results may be inaccurate.\n");
+		}
+		if (P.height > 90) {
+			fprintf(stderr, "WARNING: Height of bounding box > 90 degrees.  Results may be inaccurate.\n");
+		}
 		s = 1;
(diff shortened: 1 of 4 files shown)


src/SetupModel.cpp
```diff
@@ -184,7 +191,12 @@ void SetupModel(char* DensityFile, char* NetworkFile, char* SchoolFile, char* Re
 		P.LocationInitialInfection[i][0] -= P.SpatialBoundingBox[0];
 		P.LocationInitialInfection[i][1] -= P.SpatialBoundingBox[1];
 	}
+	// Find longest distance - may not be diagonally across the bounding box.
 	t = dist2_raw(0, 0, P.width, P.height);
+	double tw = dist2_raw(0, 0, P.width, 0);
```
igfoo, line 196: I'm not sure exactly what calculation is being done here, but does it need to check `dist2_raw(0, P.height, P.width, P.height);` too, for the other hemisphere? And does it need to do something different for regions that span the equator?

matt-gretton-dann: The `dist2_raw` is calculating the square of the great circle distance between two points (see src/Dist.cpp dist2UTM - [this](https://www.movable-type.co.uk/scripts/latlong.html) is a useful website that gives the formula and what's going on.

What the model does is read in the geography and then rotate it so that the bottom left of the bounding box is at (0 latitude, 0 longitude).  
 This is fine because the Earth is modelled as a perfect sphere.  Therefore, we don't have to worry about countries that straddle the equator, or about `dist2_raw(0, P.height, P.width, P.height)` as that will be less than `dist2_raw(0, 0, P.width, 0).  What we do possibly have to do is alert on countries that are more than 180 degrees wide or 90 degrees tall.

PR #252, opened 2020-05-13
1. Change from C to C++ struct definitions

zebmason:
2. Remove unused fields which has implications for previously recorded snapshot files

Hopefully point 2 is not an issue...

covid-sim source code provided a nice test for my struggles with libclang earlier today, hence this commit

### DIFF
--- a/src/Bitmap.cpp
+++ b/src/Bitmap.cpp
@@ -211,7 +211,7 @@ void OutputBitmap(int tp)
 	    ERR_CRITICAL_FMT("Unable to open bitmap file %s (%d): %s\n", buf, errno, errMsg);
 	  }
 	  fprintf(dat, "BM");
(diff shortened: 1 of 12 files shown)


zebmason:
@matt-gretton-dann What's the coding style for braces after a class / struct? [Allman](https://en.wikipedia.org/wiki/Indentation_style#Allman_style)?

matt-gretton-dann:
> @matt-gretton-dann What's the coding style for braces after a class / struct? [Allman](https://en.wikipedia.org/wiki/Indentation_style#Allman_style)?

@zebmason: Allman.  Or to be specific:

```cpp
class Foo
{
};
```

I will generate a ClangFormat when I have a moment which will reflect this.  I think we're going to be Allman everywhere as that is what the code has historically followed.

zebmason:
@matt-gretton-dann I assumed that it would be Allman so I implemented that in the update. Looks ready to merge.

PR #254, opened 2020-05-14
Allow passing of any extra flags to integration-test.py.

zebmason:
e.g. --covidsim ..\out\build\x64-Release\src\CovidSim.exe
allows me to run the regression tests on Windows


zebmason:
Thanks, it's just that on Windows I have different shells with different capabilities and none with the required combination.

PR #255, opened 2020-05-14
Cleanup Kernel

CloneDeath:
Some minor Kernel cleanup.

### DIFF
--- a/src/Kernels.cpp
+++ b/src/Kernels.cpp
@@ -24,7 +24,6 @@
 double *nKernel, *nKernelHR;
 void InitKernel(int DoPlaces, double norm)
 {
-	int i, j;
 	double(*Kernel)(double);
 
 	if (P.KernelType == 1)
@@ -43,19 +42,19 @@ void InitKernel(int DoPlaces, double norm)
 		Kernel = PowerExpKernel;
 	else
 		ERR_CRITICAL_FMT("Unknown kernel type %d.\n", P.KernelType);
-#pragma omp parallel for private(i) schedule(static,500) //added private i
-	for (i = 0; i <= P.NKR; i++)
+#pragma omp parallel for schedule(static,500)
+	for (int i = 0; i <= P.NKR; i++)
 	{
 		nKernel[i] = (*Kernel)(((double)i) * P.KernelDelta) / norm;
 		nKernelHR[i] = (*Kernel)(((double)i) * P.KernelDelta / P.NK_HR) / norm;
 	}
 
-#pragma omp parallel for schedule(static,500) private(i,j)
-	for (i = 0; i < P.NCP; i++)
+#pragma omp parallel for schedule(static,500)
+	for (int i = 0; i < P.NCP; i++)
 	{
 		cell *l = CellLookup[i];
 		l->tot_prob = 0;
-		for (j = 0; j < P.NCP; j++)
+		for (int j = 0; j < P.NCP; j++)
 		{
 			cell *m = CellLookup[j];
 			l->tot_prob += (l->max_trans[j] = (float)numKernel(dist2_cc_min(l, m))) * m->n;
--- a/src/Kernels.cpp
+++ b/src/Kernels.cpp
@@ -42,14 +42,17 @@ void InitKernel(double norm)
 		Kernel = PowerExpKernel;
 	else
 		ERR_CRITICAL_FMT("Unknown kernel type %d.\n", P.KernelType);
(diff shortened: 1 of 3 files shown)


src/Kernels.cpp
```diff
@@ -44,22 +43,25 @@ void InitKernel(int DoPlaces, double norm)
 		Kernel = PowerExpKernel;
 	else
 		ERR_CRITICAL_FMT("Unknown kernel type %d.\n", P.KernelType);
-#pragma omp parallel for private(i) schedule(static,500) //added private i
-	for (i = 0; i <= P.NKR; i++)
+
+#pragma omp parallel for schedule(static,500) default(none) \
+		shared(P, Kernel, nKernel, nKernelHR, norm)
+	for (int i = 0; i <= P.NKR; i++)
```
matt-gretton-dann, line 49: This is a non-obvious change.  However, I am happy about it:
 * `i` is the loop variable so automatically private
 * `default(none)` ensures we explicitly label the data-sharing of all variables.
 * All other variables are shared as they should be.

src/Kernels.cpp
```diff
@@ -44,22 +43,25 @@ void InitKernel(int DoPlaces, double norm)
...
 	{
 		nKernel[i] = (*Kernel)(((double)i) * P.KernelDelta) / norm;
 		nKernelHR[i] = (*Kernel)(((double)i) * P.KernelDelta / P.NK_HR) / norm;
 	}
 
-#pragma omp parallel for schedule(static,500) private(i,j)
-	for (i = 0; i < P.NCP; i++)
+#pragma omp parallel for schedule(static,500) default(none) \
+		shared(P, CellLookup)
+	for (int i = 0; i < P.NCP; i++)
```
matt-gretton-dann, line 57: Ditto, as above.

src/Kernels.cpp
```diff
@@ -44,22 +43,25 @@ void InitKernel(int DoPlaces, double norm)
...
 	{
 		cell *l = CellLookup[i];
 		l->tot_prob = 0;
-		for (j = 0; j < P.NCP; j++)
+		for (int j = 0; j < P.NCP; j++)
 		{
 			cell *m = CellLookup[j];
-			l->tot_prob += (l->max_trans[j] = (float)numKernel(dist2_cc_min(l, m))) * m->n;
+			l->tot_prob += (l->max_trans[j] = (float)numKernel(dist2_cc_min(l, m))) * (float)m->n;
```
matt-gretton-dann, line 64: If we're tidying code up can we split this into two statements:

```cpp
l->max_trans[j] = ...
l->tot_prob += ...
```

src/Kernels.cpp
```diff
@@ -1,6 +1,5 @@
 #include <cmath>
-#include <stdio.h>
-#include <cstddef>
+#include <iostream>
```
matt-gretton-dann, line 2: I don't see any use of <iostream> provided features here - do we need it at all?

src/Kernels.cpp
```diff
@@ -1,6 +1,5 @@
 #include <cmath>
```
matt-gretton-dann, line 1: I notice that we don't use `std::` math functions only ones in the global namespace.  We should either change the header or use the `std::` namespace explicitly.

src/Kernels.cpp
CloneDeath: It's used by `numKernel` on line 117

```
if (t > P.NKR)
{
	fprintf(stderr, "** %lg  %lg  %lg**\n", r2, P.KernelDelta, t);
	ERR_CRITICAL("r too large in NumKernel\n");
}
```

I considered changing it to modern c++ style streams with `<<` pipes, but ultimately decided to address that in a future (and separate) pull request, mostly because this PR was starting to get large, and I didn't want a cstream change to distract from my `default(none)` change above.

src/Kernels.cpp
CloneDeath: Agreed. I think it should be part of a separate PR, though.

src/Kernels.cpp
matt-gretton-dann: When we move to iostream we should do it everywhere at once.

src/Kernels.cpp
matt-gretton-dann: Agreed.

src/Kernels.cpp
CloneDeath: Split.

PR #256, opened 2020-05-14
Replaced header guards with #pragma once.

CloneDeath:
In response to #204 :

@zdroid picked the easier solution of using the header guards, since those were used more consistently, and he wasn't sure how others felt about pragma once.

However, zdroid and another developer expressed that they'd prefer pragma once, and I agree, so I went through all of the headers, and replaced the header guards.

### DIFF
--- a/src/BinIO.h
+++ b/src/BinIO.h
@@ -1,2 +1,4 @@
+#pragma once
+
 size_t fwrite_big(void *,size_t ,size_t , FILE *);
 size_t fread_big(void *,size_t ,size_t , FILE *);
--- a/src/Bitmap.h
+++ b/src/Bitmap.h
@@ -1,5 +1,4 @@
-#ifndef COVIDSIM_BITMAP_H_INCLUDED_
-#define COVIDSIM_BITMAP_H_INCLUDED_
+#pragma once
 
 #include <cstdint>
 
@@ -46,5 +45,3 @@ void OutputBitmap(int);
 void InitBMHead();
 
 void Bitmap_Finalise();
-
-#endif // COVIDSIM_BITMAP_H_INCLUDED_
--- a/src/CalcInfSusc.h
+++ b/src/CalcInfSusc.h
@@ -1,5 +1,4 @@
-#ifndef COVIDSIM_CALCINFSUSC_H_
-#define COVIDSIM_CALCINFSUSC_H_
+#pragma once
 
 double CalcHouseInf(int, unsigned short int);
 double CalcPlaceInf(int, int, unsigned short int);
@@ -9,5 +8,3 @@ double CalcHouseSusc(int, unsigned short int, int, int);
 double CalcPlaceSusc(int, int, unsigned short int, int, int);
 double CalcSpatialSusc(int, unsigned short int, int, int);
 double CalcPersonSusc(int, unsigned short int, int, int);
-
-#endif // COVIDSIM_CALCINFSUSC_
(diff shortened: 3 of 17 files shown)


zdroid:
Thanks for the change. I hope I haven't just made the task harder for you. :laughing:

zebmason:
In the new files that I've been writing I've used #pragma once. Had an argument with someone a few years back as they didn't like it because it was non-standard and my response was that it was supported by all the compilers we were targeting. Did think of stripping the include guards out myself but didn't feel like taking the punishment so good of you to do it.

matt-gretton-dann:
> In the new files that I've been writing I've used #pragma once. Had an argument with someone a few years back as they didn't like it because it was non-standard and my response was that it was supported by all the compilers we were targeting. Did think of stripping the include guards out myself but didn't feel like taking the punishment so good of you to do it.

Oh dear...  I'm going to be duplicate that annoying person :-).

I don't like this because it is not standards conformant.  However, I do take the point that all the compilers we are likely to support permits `#pragma once`.

Let me think on this.

CloneDeath:
@matt-gretton-dann Any new thoughts?

We're starting to create new classes, and with them new headers.

Here's my argument (for what it's worth):
`#pragma once` is pretty popular (among compilers and developers), and it's not like we're locking ourselves into one OS, or using some esoteric feature. Worst case scenario, we *can* go back to header guards if needed. But I rather have the convenience of `#pragma once` over worrying about some future compiler that may not ever exist.

Plus, like I said, it's not an esoteric invasive feature (like `async` in js. Which I love, but it's definitely invasive and spreads like a virus... but that's neither here nor there). `#pragma once` is a mild convenience, and can be easily backed out of (as opposed to switching from .then to async/await in js).

Lastly, it keeps getting proposed for standardization. The main reason it hasn't been accepted (from what I understand) is a lack of consensus on how to identify a file as unique. I've heard other arguments too, that's not the only one. But the general consensus seems to be that people want it standardize, and I believe it will be eventually.

I could go on, but maybe if you explained your reservations some more, we could help ease them?

weshinsley:
Quick thoughts on this one - (4) makes me feel we should probably action something.

1. I don't think the old-school #ifndef _HEADER #define _HEADER .... #endif pattern is particularly ugly - although #pragma once is a neater one-liner for sure.

2. The only reservations (on my part) come from wanting to be abudently cautious with standards. I wouldn't want some squawking of "the code doesn't even compile on all platforms" to arise, should someone choose an outlandish compiler to make their case with.

3. However... does anyone know of any compiler that would compile the current code, yet would fail to do so if we used `#pragma once`? 

4. I also note we seem to have a couple of `#pragma once` already in `Direction.hpp` and `MicroCellPosition.hpp` - (which are also our first `.hpp` files). So I think we should probably be consistent one way or the other.

zdroid:
@weshinsley 4) Those were created by @CloneDeath. I also prefer `.hpp` over `.h` since it makes clear what language is it written in (and GitHub recognizes them as C++, instead of C).

About pragma once support, check Portability on https://en.wikipedia.org/wiki/Pragma_once.

CloneDeath:
Rebased to fix merge conflict with deleted file.

zebmason:
@weshinsley  3) There are countless compilers out there and thus plenty that are bound not to work, especially historical versions. It is always possible for some naysayer to fork the codebase and replace the pragma once's with include guards so I don't see this as being an issue. Bit of a pain for them to always be rebasing their fork but do you actually care?

weshinsley:
With apologies to those who will find this a bit boring... 
closing this in preference for https://github.com/mrc-ide/covid-sim/pull/443

PR #258, opened 2020-05-15
Add GoogleTest support

zebmason:
If this works on CI then the build output will include the results of successfully running a test


CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 # CMakeLists.txt
 
 # CMake setup
-cmake_minimum_required (VERSION 3.8)
+cmake_minimum_required (VERSION 3.11)
```
matt-gretton-dann, line 4: I'm guessing this is one of the issues with the scripts.  At least the `ubuntu-latest` images are with CMake 3.10.

If this is the issue - then I'm happy with a CMake version bump as long as you update the install_dependencies.sh scripts to fix the CI loop.

CMakeLists.txt
```diff
@@ -47,11 +47,13 @@ endif()
...
 
+add_subdirectory(third_party)
 add_subdirectory(src)
 
+add_subdirectory(unit_tests)
 if(WIN32 AND USE_WIN32_BMP)
-  message(WARNING "USE_WIN32_BMP is enabled.  Therefore testing is disabled.")
-  message(WARNING "To enable testing invoke with `cmake -DUSE_WIN32_BMP=NO`")
+  message(WARNING "USE_WIN32_BMP is enabled.  Therefore regression testing is disabled.")
+  message(WARNING "To enable regression testing invoke with `cmake -DUSE_WIN32_BMP=NO`")
```
matt-gretton-dann, line 56: Oh - that's a bug.  In fact `USE_WIN32_BMP` is no longer relevant, as we have the `/BM:fmt` command line option.

So don't worry about that change - I'll do something to drop it.

third_party/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+  GIT_REPOSITORY https://github.com/google/googletest.git
+  GIT_TAG        release-1.10.0
+)
+FetchContent_GetProperties(googletest)
+if(NOT googletest_POPULATED)
+  FetchContent_Populate(googletest)
+  set(gtest_force_shared_crt ON CACHE BOOL "" FORCE) 
+  add_subdirectory(${googletest_SOURCE_DIR} ${googletest_BINARY_DIR})
+endif()
+include(GoogleTest)
```
matt-gretton-dann, line 14: Missing NL at EOF.

unit_tests/noddy/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
...
+add_executable(test2 Holder.cpp)
+target_link_libraries(test2 gtest_main ${CMAKE_THREAD_LIBS_INIT})
+add_test(test2 test2)
+
+add_custom_target(test2_run
+  COMMAND test2
+  WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+)
+
+add_dependencies(test1 test2_run)
```
matt-gretton-dann, line 19: Missing NL at EOF

visual_studio/test-noddy/test-noddy.vcxproj
```diff
@@ -0,0 +1,103 @@
...
+  </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="covid-sim.vcxproj">
+      <Project>{1284b4ee-852b-4bb5-b6ef-60d8d9eb46dc}</Project>
+    </ProjectReference>
+  </ItemGroup>
+  <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
+  <ImportGroup Label="ExtensionTargets">
+  </ImportGroup>
+</Project>
```
matt-gretton-dann, line 103: Missing NL at EOF.

.gitignore
```diff
@@ -13,9 +12,9 @@ build/
 build-*/
 build_*/
 .project
-*.vcxproj.user
 tests/*-output/
 .idea/
+/CMakeSettings.json
```
matt-gretton-dann, line 17: This is an interesting one.  I initially thought about adding this a while ago.  But then I decided against it as the contents of the file are not specific to the location on a drive, and so having an initial set of builds committed may be useful.

I am going to revisit what the issues are with CMake because I would like to get to one build system and the other changes to handle these are painful.  (I would consider dropping the regression tests from the Visual Studio project and just documenting how to run them from the command line).

CMakeLists.txt
zebmason: What do you think of https://github.com/marketplace/actions/get-cmake ?

CMakeLists.txt
zebmason: Reverted

third_party/CMakeLists.txt
zebmason: Added

unit_tests/noddy/CMakeLists.txt
zebmason: Added

visual_studio/test-noddy/test-noddy.vcxproj
zebmason: Added

zebmason:
@matt-gretton-dann I could comment out adding the third_party and unit_tests subdirectories to make the code safe for CI and allow developers to uncomment.

.gitignore
zebmason: Would you like me to revert this change?

zebmason:
Oh, and revert the cmake version as well

.gitignore
matt-gretton-dann: Yes please.  Sorry for not being clear above.

CMakeLists.txt
matt-gretton-dann: I don't object - its a public repo action with a suitable license.  As long as you configure it for the version of cmake shipped in Windows Visual Studio (3.16.x I think).  Ubuntu 20.04 and macOS via brew both ship 3.17.

But my further question is: Will it work with the docker container runs?

.gitignore
zebmason: Done

CMakeLists.txt
zebmason: Couldn't get that action to work so tried something that should have been easy. I'm stumped as to why Ubuntu bionic is failing - it is setting the path but then apparently ignoring it.

Still getting used to this CI, I thought that you would have to set it up as I am familiar with Hudson/Jenkins and latterly Travis.

zebmason:
Ubuntu bionic after adding the downloaded CMake to the path:
 build0s
##[error]Process completed with exit code 1.
Run mkdir build && cd build
CMake Error at CMakeLists.txt:4 (cmake_minimum_required):
  CMake 3.11 or higher is required.  You are running version 3.10.2

CMakeLists.txt
insidedctm: Am I right in thinking that all the other containers ship with 3.17 and only ubuntu:bionic ships with 3.10? So the manual install of 3.17 is only relevant for ubuntu:bionic? Is it possible that the export is not being picked up because the subsequent processes are separate from the process running install_dependencies.sh?

zebmason: @insidedctm Most of the containers were failing until I added the download of CMake (tar -xf and add to path) after which it was only bionic. So I've commented out traversing into third_party and unit_tests and rolled back the required CMake version. I know when I am beat.

PR #262, opened 2020-05-15
Strip trailing whitespace

zdroid:
All this extraneous whitespace produces unwanted diffs and thus makes it impossible for me to work in VSCode or basically any modern editing software properly.


zdroid:
@matt-gretton-dann done

PR #264, opened 2020-05-16
Cleanup parallel for loops

CloneDeath:
Focused on specifying the shared variables for all the parallel for loops, and reducing the number of explicit private variables by moving them into scope.

I haven't removed all the private variables, as some of them are a bit more complicated.

I also haven't touched Kernal.cpp to avoid conflicts with #255 (my other PR)


CloneDeath:
Working on fixing build issues for other platforms.

CloneDeath:
Compilation issues fixed.

It looks like the way that stderr expands on some OS (windows and alpine) conflicted with using stderr inside of a shared specifier. (unrelated, but I am currently on Ubuntu, so I wasn't aware of this issue until the CI ran).

To solve this, I assigned stderr to a temporary variable stderr_shared and specified that stderr_shared as a shared variable (that doesn't expand), preventing the invalid expansion for those two OS.

It's not a pretty solution, but I didn't want to ALSO re-engineer how errors are printed in this PR. I'd like to do that in a separate followup PR, after this is merged.

src/SetupModel.cpp
```diff
@@ -1112,9 +1112,10 @@ void SetupPopulation(char* DensityFile, char* SchoolFile, char* RegDemogFile)
 	if (!(Households = (household*)malloc(P.NH * sizeof(household)))) ERR_CRITICAL("Unable to allocate household storage\n");
 	for (j = 0; j < NUM_AGE_GROUPS; j++) AgeDist[j] = AgeDist2[j] = 0;
 	if (P.DoHouseholds) fprintf(stderr, "Household sizes assigned to %i people\n", numberOfPeople);
+
+	FILE* stderr_shared = stderr;
```
matt-gretton-dann, line 1116: This is ugly - but the correct fix I think for the current code usage.

CloneDeath:
Yes. Since there were some conflicts, I did a merge instead of a rebase. Let me know if you would rather I rebase instead, and I can do that quickly for you.

PR #265, opened 2020-05-16
Remove MachineDefines.h

zdroid:
This is a move towards more idiomatic file structure. Currently the linking of source files is a mess, and removing one that's not needed will reduce the mess.


src/Constants.h
```diff
@@ -63,6 +63,9 @@ constexpr double EARTH_DIAMETER = EARTH_CIRCUMFERENCE / PI;
  */
 constexpr double EARTHRADIUS = EARTH_DIAMETER / 2;
 
+#define MAX_NUM_THREADS 96
+#define CACHE_LINE_SIZE 64
+
```
matt-gretton-dann, line 68: As you're moving these - make them into `const int` please.

zdroid: Done

zdroid: ping @matt-gretton-dann

weshinsley:
@matt-gretton-dann - also we should update .vcxproj file when this is merged.

zdroid:
@weshinsley that change already is in this PR

weshinsley:
Requested changes were addressed - simple changed - resolved and merged.

zdroid:
@weshinsley Great work! You may want to use squash or rebase instead of merge, as it doesn't make a merge commit.

weshinsley:
Ooh right - I'm not sure how to do that from your branch -  I've been strictly following the command-line instructions when github can't do the conflict resolution on the page...

zdroid:
@weshinsley I deleted my repository along with this and the other branches, so it probably made the job harder for you. My last 3 PRs should be fine regarding that.

PR #268, opened 2020-05-16
Everywhere else P.SpatialBoundingBox is asked if the point is inside …

zebmason:
…with
inequality on the upper bound. So asking for outside should be >= rather than
merely > for consistency.

Is this a correct interpretation @NeilFerguson ?


matt-gretton-dann:
Thank you for this - to repeat a comment from @dlaydon:  This does not affect published results as the airport functionality has not been used for the COVID-19 simulations

PR #269, opened 2020-05-16
Experiment to see how CI copes with C++ 17 standard.

zebmason:
Note that gcc does not fully support the standard library for this standard.

@cjgdev Christopher thought you might be interested

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,7 +45,7 @@ else()
 endif()
 
 # specify the C++ standard
-set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD 17)
 
 add_subdirectory(src)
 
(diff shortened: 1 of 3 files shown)


zebmason:
Sorry forgot to set it to draft, still getting used to collaborating on GitHub rather than on one of your competitors. This was a test given the comment in #207 by @cjgdev.

I used `std::filesystem` purely as a test. I got caught out on some consultancy I did last year where I found that when the client supplied me with the Linux API for stage 2 that the supported version of gcc didn't do C++17 and I had to use a [back port](https://github.com/gulrak/filesystem) to C++14. Once CMake is upgraded I'll raise an issue to pull that in and do the PR myself unless the reviewers object.

zebmason:
Have example code for backport of filesystem in #289

PR #270, opened 2020-05-17
Local variable for density file magic number

galactic-src:
A couple of minor improvements (but every little helps, right?)
- remove the 2 parameters to InitKernel, neither of which is really used
- create a separate int for density file magic number detection, rather than confusingly reusing P.BinFileLen


galactic-src:
@matt-gretton-dann you are welcome, and done. The whitespace should also hopefully be sorted now.

PR #273, opened 2020-05-17
Replace InfStat.h defines with enumerators

zdroid:


dlaydon:
Hi @zdroid, thanks for this. We've just updated the regression tests, which I strongly suspect is the only reason your changes haven't passed the checks. If you can update this branch with new regression tests (and assuming your changes then pass), then I'll approve. Thanks!

dlaydon:
Thanks @zdroid

PR #283, opened 2020-05-18
Replace ERR_CRITICAL(_FMT) with ErrorCritical

zdroid:
The function can be greatly simplified, since it requires only one argument, plus ellipsis, compared to the 3 arguments currently required. It also avoids using macros for functions which is a bad practice.

### DIFF
--- a/src/Bitmap.cpp
+++ b/src/Bitmap.cpp
@@ -155,7 +155,7 @@ void OutputBitmap(int tp)
 	  using namespace Magick;
 	  fprintf(stderr, "\noutputing ImageMagick stuff");
 	  sprintf(buf, "%s.bmp", OutF);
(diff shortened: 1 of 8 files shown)


igfoo:
Thank you for the pull request! However, unfortunately this will make all errors be reported as being from Error.cpp line 9, rather than the location of the ERR_CRITICAL call.

zdroid:
@igfoo Wouldn't moving `ErrorCritical`'s declaration into `Error.h` (i.e. make it header-only) solve that issue? `__FILE__` and `__LINE__` supposedly resolve to `.cpp` and not `.h`.

igfoo:
I'm not sure I follow, but feel free to open a PR and we can take a look. You should be able to test by making `main` immediately call the error function.

PR #285, opened 2020-05-19
correct data type for MAX_DIST_AIRPORT_TO_HOTEL const

insidedctm:
In [#280](https://github.com/mrc-ide/covid-sim/pull/280) MAX_DIST_AIRPORT_TO_HOTEL was converted from #DEFINE to const int. I believe this should have been const double.


zdroid:
:+1: I edited that one manually and still made a mistake. :smile:

insidedctm:
> 👍 I edited that one manually and still made a mistake. 😄

So easy to do :)

PR #286, opened 2020-05-19
add individual state level admin files

jhelie:
This adds the state level admin files that are generated on the fly by the UI. @gorzell I've noticed that the `[Population size]` is always equal to the total US pop size (`338798532`): is it something that should be changed?
(for the few files which were already in the repo and that I've simply moved, such as `Alaska_admin.txt` or `Hawaii_admin.txt`, the number seems to reflect the actual population of the admin unit).


weshinsley:
@matt-gretton-dann - this PR fell asleep - I've moved the USA files back into the top level.

@dlaydon - I can't remember whether `[Population size]` for these state level files should be USA population, or state specific; it'll be loading from the same population file...

matt-gretton-dann:
> @matt-gretton-dann - this PR fell asleep - I've moved the USA files back into the top level.
> 
> @dlaydon - I can't remember whether `[Population size]` for these state level files should be USA population, or state specific, or whether it matters - it'll be loading the different states out of the same wpop file... (except for the non "mainland-USA" - Alaska, Hawaii, Guam, Virgin Islands US, Puerto Rico, which have their [Population size] set specifically already)
> 
> The more I think about it, the more I think this is correct; that the states dug out of the "all-US" wpop file (with the extra schools data) should have the Population Size set to USA, and the others shouldn't - which is what we have... but would be good to confirm

How does the `[Fix population size at specified value]` setting interact with the population size?

I need to update stuff the run_sample.py scripts to understand which wpop files to use for these files.

matt-gretton-dann:
I have retargeted this to mainline as well (it was targeted at one of my own branches).

PR #287, opened 2020-05-19
Replace math.h with cmath

zdroid:
It's the preferred modern C++ header.

Also replaces `(double)abs(double)` with `std::fabs(double)` in `Dist.cpp` since it's explicitly double.

### DIFF
--- a/src/Bitmap.cpp
+++ b/src/Bitmap.cpp
@@ -2,7 +2,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
-#include <math.h>
+#include <cmath>
 
 #include "BinIO.h"
 #include "Bitmap.h"
--- a/src/CalcInfSusc.cpp
+++ b/src/CalcInfSusc.cpp
@@ -1,4 +1,4 @@
-#include <math.h>
+#include <cmath>
 
 #include "CalcInfSusc.h"
 #include "Constants.h"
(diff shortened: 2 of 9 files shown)


zdroid:
I really have no clue why macOS build constantly fails on ambiguous `std::abs`. Makes me wonder does it even have C++ headers implemented properly. :confused:

CloneDeath:
Honestly, I think the fact that they are shoving two states into one enum, and encoding it as a combination of the integer value and it's sign, is a problem. I'd like to pull that out into two separate fields in the future...

But back on topic:
I think it's fine to just explicitly use one of the predefined std:abs instead of relying on the template. Just case it to a double instead of an int, is my vote. It sucks, but I think it's still a net move in the positive direction. Plus, with my tangent above, it should ideally be removed entirely one day.

matt-gretton-dann:
I think the macOS problem is that because we are compiling for C++14 the `std::abs(int)` overload is not made visible by the `<cmath>` header - you need to include <cstdlib> as well to get the overloads to work.

CloneDeath:
I ended up including cstdlib in my other PR. #292

https://stackoverflow.com/questions/60463014/cmath-error-on-macos-c17-call-to-abs-ambiguous

zdroid:
@matt-gretton-dann https://en.cppreference.com/w/cpp/numeric/math/abs, https://en.cppreference.com/w/cpp/numeric/math/fabs
`<cmath>` is the only header that includes both in C++14.

@CloneDeath I guess that enum can be reworked to avoid integer values altogether, but that's a thing for another PR.

zdroid:
It failed after that fix as well. @CloneDeath

PR #288, opened 2020-05-19
Add status badges for the workflows

zebmason:


zebmason:
Status seemed to be the most appropriate. I got bored of messing with the googletest branch so decided to have a play.

PR #289, opened 2020-05-19
Use a back port of std::filesystem

zebmason:
Not actually sure that this needed anymore. Admittedly error statuses are checked on fopen which may fail because a file either does or doesn't exist. Would be better to use filesystem::exists in that case.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 # CMakeLists.txt
 
 # CMake setup
-cmake_minimum_required (VERSION 3.8)
+cmake_minimum_required (VERSION 3.11)
 
 # Project initialisation
 project("CovidSim")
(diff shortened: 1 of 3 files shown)


matt-gretton-dann:
I think for all the compilers we care about the following macro hack may work:
```cpp
#ifdef __cpp_lib_filesystem
#include <filesystem>
using namespace fs = std::filesystem;
#elif defined(__cpp_lib_experimental_filesystem)
#include <experimental/filesystem>
using namespace fs = std::experimental::filesystem;
#else
#error "Unable to find filesystem header."
#endif
```

It does mean we use an `fs` namespace and not the `std` one - but I don't think this should be an issue.

weshinsley:
(Ref Issue #290)

PR #292, opened 2020-05-19
Abstract inf state in person into several functions.

### LINKED ISSUE #291: Add function to check if person is dead or alive
I see a lot of code that looks like `abs(Hosts[j].inf) != InfStat_Dead`.

I'd like to introduce `isDead()` and `isAlive()` functions to person in order wrap this check.

Edit: I wanted to assign this ticket to myself, as I was planning on doing this tonight, but it looks like I don't have permissions.

(honestly, this ticket is so I don't forget).

CloneDeath:
Added Person.cpp with persion function definitions.
Replaced all alive and dead checks with isAlive and isDead.

Fixes #291

### DIFF
--- a/src/CovidSim.cpp
+++ b/src/CovidSim.cpp
@@ -2987,7 +2987,7 @@ int RunModel(int run) //added run number as parameter
 	for (i = j = k = ni = fs2 = i2 = 0; i < P.N; i++)
 	{
 		if (i % 1000 == 0) fprintf(stderr, "\r*** %i              ", i);
-		if (Hosts[i].inf == 0) j++;
+		if (Hosts[i].infectionState == 0) j++;
 		if ((Hosts[i].pcell < P.NC) && (Hosts[i].pcell >= 0))
 		{
 			if (Cells[Hosts[i].pcell].susceptible[Hosts[i].listpos] != i)
@@ -3000,7 +3000,7 @@ int RunModel(int run) //added run number as parameter
 			else
 			{
 				if ((Hosts[i].listpos > Cells[Hosts[i].pcell].S - 1) && (Hosts[i].is_susceptible())) i2++;
-				if ((Hosts[i].listpos < Cells[Hosts[i].pcell].S + Cells[Hosts[i].pcell].L + Cells[Hosts[i].pcell].I - 1) && (abs(Hosts[i].inf) == InfStat_Recovered)) i2++;
+				if ((Hosts[i].listpos < Cells[Hosts[i].pcell].S + Cells[Hosts[i].pcell].L + Cells[Hosts[i].pcell].I - 1) && (abs(Hosts[i].infectionState) == InfStat_Recovered)) i2++;
 			}
 			if ((Cells[Hosts[i].pcell].S + Cells[Hosts[i].pcell].L + Cells[Hosts[i].pcell].I + Cells[Hosts[i].pcell].R + Cells[Hosts[i].pcell].D) != Cells[Hosts[i].pcell].n)
 			{
@@ -3134,7 +3134,7 @@ int RunModel(int run) //added run number as parameter
 		for (i = j = k = NumSeedingInfections = fs2 = i2 = 0; i < P.PopSize; i++)
 		{
 			if (i % 1000 == 0) fprintf(stderr, "\r*** %i              ", i);
-			if (Hosts[i].inf == 0) j++;
+			if (Hosts[i].infectionState == 0) j++;
 			if ((Hosts[i].pcell < P.NC) && (Hosts[i].pcell >= 0))
 			{
 				if (Cells[Hosts[i].pcell].susceptible[Hosts[i].listpos] != i)
@@ -3147,7 +3147,7 @@ int RunModel(int run) //added run number as parameter
 				else
 				{
 					if ((Hosts[i].listpos > Cells[Hosts[i].pcell].S - 1) && (Hosts[i].is_susceptible())) i2++;
-					if ((Hosts[i].listpos < Cells[Hosts[i].pcell].S + Cells[Hosts[i].pcell].L + Cells[Hosts[i].pcell].I - 1) && (abs(Hosts[i].inf) == InfStat_Recovered)) i2++;
+					if ((Hosts[i].listpos < Cells[Hosts[i].pcell].S + Cells[Hosts[i].pcell].L + Cells[Hosts[i].pcell].I - 1) && (abs(Hosts[i].infectionState) == InfStat_Recovered)) i2++;
 				}
 				if ((Cells[Hosts[i].pcell].S + Cells[Hosts[i].pcell].L + Cells[Hosts[i].pcell].I + Cells[Hosts[i].pcell].R + Cells[Hosts[i].pcell].D) != Cells[Hosts[i].pcell].n)
 				{
--- a/src/CovidSim.cpp
+++ b/src/CovidSim.cpp
@@ -3076,7 +3076,7 @@ int RunModel(int run) //added run number as parameter
 								do
 								{
 									l = (int)(((double)P.PopSize) * ranf()); //// choose person l randomly from entire population. (but change l if while condition not satisfied?)
-								} while (Hosts[l].isDead() || (ranf() > P.FalsePositiveAgeRate[HOST_AGE_GROUP(l)]));
+								} while (Hosts[l].is_dead() || (ranf() > P.FalsePositiveAgeRate[HOST_AGE_GROUP(l)]));
 								DoFalseCase(l, t, ts, 0);
 							}
 						}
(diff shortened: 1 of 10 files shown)


CloneDeath:
This PR fixes #291

CloneDeath:
I'm using cstdlib instead of cmath because of a bug in abs on mac.

https://stackoverflow.com/questions/60463014/cmath-error-on-macos-c17-call-to-abs-ambiguous

CloneDeath:
Looks like build failed on ubuntu because of environment setup error (couldn't reach server to download packages), not because of code error.

I am running ubuntu, and it works for me.

src/Model.h
```diff
@@ -12,7 +12,7 @@
 
 #pragma pack(push, 2)
 
-typedef struct PERSON {
+struct person {
```
matt-gretton-dann: I've just merged #252 which does this globally and calls it `struct Person`.  You'll need to resolve these conflicts.

src/Model.h
```diff
@@ -50,7 +51,9 @@ typedef struct PERSON {
 	unsigned short int dct_start_time, dct_end_time, dct_trigger_time, dct_test_time; //digital contact tracing start and end time: ggilani 10/03/20
 	int ncontacts; //added this in to record total number of contacts each index case records: ggilani 13/04/20
 
-} person;
+	bool isAlive() const;
+	bool isDead() const;
```
matt-gretton-dann, line 55: Can we use snake_case for method names please?  (so is_alive() and is_dead()).

CloneDeath: No problem. Done. :smile:

src/CovidSim.cpp
```diff
@@ -2964,7 +2964,7 @@ int RunModel(int run) //added run number as parameter
 	for (i = j = k = ni = fs2 = i2 = 0; i < P.N; i++)
 	{
 		if (i % 1000 == 0) fprintf(stderr, "\r*** %i              ", i);
-		if (Hosts[i].inf == 0) j++;
+		if (Hosts[i].infectionState == 0) j++;
```
CloneDeath, line 2967: Code is commented out.

Either we should uncomment it, or delete it. I'll save that for a future PR.

CloneDeath:
@matt-gretton-dann  I went ahead and made the variable private, and wrapped all the logic into new functions.

That single enum state really needs to be multiple variables, but I wanted to focus on refactoring out the usage of the variable without changing the logic, before we tackled that.

CloneDeath:
Rebased to fix merge conflicts.

weshinsley:
Hmm - I think PR #387 (which was just merged) might have overlapped with some things here...

mstevens-uk:
The ambition of both PRs is to raise the level of abstraction of the infection state details - PR 387 is a good start but there is a lot more to do.

src/CovidSim.cpp
```diff
@@ -5311,16 +5311,16 @@ void RecordInfTypes(void)
...
 				if (Hosts[i].latent_time * P.TimeStep <= P.SampleTime)
 					TimeSeries[(int)(Hosts[i].latent_time * P.TimeStep / P.SampleStep)].Rdenom++;
 				infcountry[Mcells[Hosts[i].mcell].country]++;
-				if (abs(Hosts[i].inf) < InfStat_Recovered)
+				if (Hosts[i].has_not_recovered())
 					l = -1;
 				else if (l >= 0)
 					l++;
-				if ((l >= 0) && ((Hosts[i].inf == InfStat_RecoveredFromSymp) || (Hosts[i].inf == InfStat_Dead_WasSymp)))
+				if (l >= 0 && ((Hosts[i].has_recovered() || Hosts[i].is_dead()) && Hosts[i].is_symptomatic()))
```
mstevens-uk, line 5323: Could this be simplified? Can we check logic has not changed from original?

src/Sweep.cpp
```diff
@@ -103,7 +103,7 @@ void TravelDepartSweep(double t)
 						l = Airports[i].DestMcells[l].id;
 						int k = (int)(ranf_mt(tn) * ((double)Mcells[l].n));
 						int i2 = Mcells[l].members[k];
-						if ((abs(Hosts[i2].inf) < InfStat_InfectiousAsymptomaticNotCase) && (Hosts[i2].inf != InfStat_Case))
+						if (Hosts[i2].is_exposed() || Hosts[i2].is_susceptible())
```
mstevens-uk, line 106: Check that this is a 1-1 with the original logic

src/Sweep.cpp
```diff
@@ -1137,16 +1136,16 @@ void DigitalContactTracingSweep(double t)
 					else if (P.FindContactsOfDCTContacts)
 					{
 						//check every day to see if contacts become index cases - but they have to be infectious. Otherwise we could set the trigger time and cause their contacts to be traced when they are not being traced themselves.
-						if ((Hosts[contact].index_case_dct == 0) && ((abs(Hosts[contact].inf) == 2) || (Hosts[contact].inf == -1)))
-							//if ((Hosts[contact].dct_test_time == ts) && (Hosts[contact].index_case_dct == 0) && ((abs(Hosts[contact].inf) == 2) || (Hosts[contact].inf == -1)))
+						if (Hosts[contact].index_case_dct == 0 && Hosts[contact].is_positive())
```
mstevens-uk, line 1139: Check that this is a 1-1 with the original logic

src/Sweep.cpp
```diff
@@ -1137,16 +1136,16 @@ void DigitalContactTracingSweep(double t)
...
 						{
 							//set them to be an index case
 							Hosts[contact].index_case_dct = 1;
 							//set trigger time to pick up their contacts in the next time step
 							Hosts[contact].dct_trigger_time = ts + 1; //added the +1 here so that if there are no delays, the contacts will still get picked up correctly
-							//if they are asymptomatic, i.e. specifically if they have inf flag 2, call DoDetectedCase in order to trigger HQ and PC too.
-							if (Hosts[contact].inf == 2)
+
+							if (Hosts[contact].is_asymptomatic())
```
mstevens-uk, line 1146: Check that this is a 1-1 with the original logic

src/Update.cpp
```diff
@@ -738,86 +738,20 @@ void DoDetectedCase(int ai, double t, unsigned short int ts, int tn)
...
 	int j, k, f, j1, j2;
-	Person* a;
 	int age;
 
 	age = HOST_AGE_GROUP(ai);
 	if (age >= NUM_AGE_GROUPS) age = NUM_AGE_GROUPS - 1;
-	a = Hosts + ai;
-	if (a->inf == InfStat_InfectiousAlmostSymptomatic) //// if person latent/asymptomatically infected, but infectious
+	Person* a = Hosts + ai;
+	if (a->is_positive() && a->is_exposed())
```
mstevens-uk, line 752: Check that this is a 1-1 with the original logic. is_positive() refers to the underlying enum details, not infection states

src/Model.h
```diff
@@ -52,6 +51,32 @@ struct Person
...
+	bool is_symptomatic() const;
+	bool is_asymptomatic() const;
+	bool is_case() const;
+	bool is_exposed() const;
+	bool is_infectious() const;
+	bool is_infected() const;
+	bool has_recovered() const;
+	bool has_not_recovered() const;
+	bool is_positive() const;
+	bool is_alive() const;
```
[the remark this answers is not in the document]
CloneDeath: is_dead returns !is_alive (or the other way around, I forget) to avoid such a scenario. Having both functions (in my opinion) improves readability.

src/CovidSim.cpp
CloneDeath: Logic has not changed from the original. I triple checked each conversion.
I believe a lot of the old logic can be simplified, but I didn't want to risk getting it wrong, and breaking things.

My approach is to first abstract the logic into functions, then split the underlying enum into separate variables, THEN simplify the logic. Simplifying the logic on a few variables is going to be a lot easier than simplifying the logic on a range on INT values.

src/Sweep.cpp
CloneDeath: All of these are 1:1 with the original logic, to maintain the same logic. This was my primary goal in this PR. This means that sometimes things look a bit "nonsensical", but it's the best description of the underlying logic I could come up given what the old logic was.

As I said before, in refactoring the enum and simplifying the equations, I hope to come to better names as well in the future, to describe the underlying logic.

src/Update.cpp
CloneDeath: These functions just wrap the inf variable for now. We're checking for one specific value, but i didn't want to make a function call for every single state (that'd defeat the purpose of trying to abstract out the logic). There are exceptions, but...

In this case, is_positive and is_exposed overlap in their coverage by a single value: InfectiousAlmostSymptomatic.

CloneDeath:
@mstevens-uk I think this PR has grown in scope much larger than originally intended.

I am going to mark this PR as Draft, and break it down into several smaller (incremental) PR, like I'm doing for the Vector math changes.

weshinsley:
Unless anyone feels differently, I suggest closing this one in preference for PR #445 - I've tried to pull in all the changes in this one, and then followed it through to completion, with docs.

PR #296, opened 2020-05-19
Changed DoVacc to return void.

### LINKED ISSUE #295: DoVacc return value is never used.
DoVacc returns an int, but the return value is never used.

Remove the return value, and make it void like all the other Update functions.

CloneDeath:
Fixes #295


weshinsley:
I think this change got implemented by #320

PR #297, opened 2020-05-19
Moved documentation comments in infstat

### LINKED ISSUE #294: Fix documenting comments for InfStat
Right now, the documenting comments for the entries in InfStat occur AFTER the value. The result is, the document comment is now describing the value AFTER the value it is intended to document.

Comments should be adjusted to be before the line of code they document.

I'm sure this happens in other places too, but I was just working in this area when I noticed it, and wanted to make a note to fix this instance later.

CloneDeath:
Fixes #294


weshinsley:
Incidentally, the original comments looked fine for spaces but not tabs...! ;-)

CloneDeath:
@weshinsley It's not just GitHub, but also the IDE. The problem is, documentation comments describe the NEXT line of code. So, if you hover over the variable elsewhere in the code, you get the wrong documentation (the documentation meant for the previous line of code is shown in intellesense).

weshinsley:
Personal preference I guess - in VS with a wide-screen, I actually find the same-line comments much clearer to read (and they are all lined up in VS - spaces/tabs here are different) - but when looking on github, I think it looks better separate lines.

CloneDeath:
Yeah, that's how I noticed the issue. I was working on some relevant code, and getting completely nonsensical documentation. This is just something we need to be cognisant of going forward.

PR #299, opened 2020-05-20
Fix doxygen build on master

matt-gretton-dann:
This PR fixes the doxygen build failures on master.  Also makes the flow run on all pushes (but only deploy on a push to master) to make checking for issues easier down the line)

(Caused by #238 and doxygen CI flow historically only being run on pushes to master - CC @zebmason).

### DIFF
--- a/.github/workflows/doxygen.yml
+++ b/.github/workflows/doxygen.yml
@@ -14,6 +14,9 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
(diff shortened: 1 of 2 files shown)


matt-gretton-dann:
Breaking the rules and force pushing this as an admin without full review as it will make the CI loop on master green.

zebmason:
It shows green now on my [badges](https://github.com/zebmason/covid-sim/tree/badges) branch

PR #302, opened 2020-05-20
Refactor CovidSim icdf array population into helper function

mstevens-uk:
A helper function setICDF() has been added to CovidSim to consolidate the population logic for icdf arrays.

### DIFF
--- a/src/CovidSim.cpp
+++ b/src/CovidSim.cpp
@@ -59,7 +59,7 @@ int GetInputParameter(FILE*, FILE*, const char*, const char*, void*, int, int, i
 int GetInputParameter2(FILE*, FILE*, const char*, const char*, void*, int, int, int);
 int GetInputParameter3(FILE*, const char*, const char*, void*, int, int, int);
 
-void setICDF(double* icdf, int startValue);
+void SetICDF(double* icdf, int startValue);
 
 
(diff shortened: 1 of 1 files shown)


src/CovidSim.cpp
```diff
@@ -59,6 +59,8 @@ int GetInputParameter(FILE*, FILE*, const char*, const char*, void*, int, int, i
 int GetInputParameter2(FILE*, FILE*, const char*, const char*, void*, int, int, int);
 int GetInputParameter3(FILE*, const char*, const char*, void*, int, int, int);
 
+void setICDF(double* icdf, int startValue);
```
CloneDeath, line 62: Function name should be SetICDF to conform with the other PascalCase global function names.

src/CovidSim.cpp
```diff
@@ -5665,3 +5646,16 @@ int GetInputParameter3(FILE* dat, const char* SItemName, const char* ItemType, v
 	//	fprintf(stderr,"%s\n",SItemName);
 	return FindFlag;
 }
+
+/* helper function to set icdf arrays */
+void setICDF(double* icdf, int startValue)
+{
+	int i;
```
CloneDeath, line 5653: Move i declaration inline to for loop.

ie: `for (int i = 0; i < CDF_RES; i++)`

src/CovidSim.cpp
mstevens-uk: Agreed - change committed.

src/CovidSim.cpp
mstevens-uk: Agreed - change committed.

PR #303, opened 2020-05-20
Use cstring, cerrno, cstddef, climits, cstdint, csignal, ctime

zdroid:
Replaces minor headers with their modern variants


weshinsley:
Have resolved the conflicts (which were just those caused by merging #301 and merged.

Issue #305, opened 2020-05-20
Use a widely-available coding style and enforce it


zdroid:
This is fairly easy to do with VS Code / Visual Studio. This project's code is currently highly unreadable and that's very detrimental to the future development. `Visual Studio` style is very similar to this project's one, among the widely-available ones.

There is also a lot of tab-space mixing which looks horrible on every setup.

CloneDeath:
Agreed. Maybe we should start a contributing guide?

ozmorph:
I agree as well.

I would personally prefer 4 spaces to tabs simply because they make GitHub Pull Requests easier to read. GitHub's code view defaults a tab to 8 spaces, which in a code base that relies on very nested logic (like this one) makes the code shift to the right so much that it sometimes wraps around.

CloneDeath:
I prefer tabs, but don't care enough to argue it.

Also, take that survey with a grain of salt, as it really is more of a reflection of what IDEs are popular and their default settings.

I say we just start with a popular style guide that sort of matches what we want the code to look like (not necessarily what it is now), and make adjustments from there.

ozmorph:
Very true about the survey and its findings. I won't argue the tabs vs. spaces debate, just stating my personal preference. I'll use whatever the community decides.

zebmason:
During a previous conversation with @matt-gretton-dann he said he'd put together a clangformat file. Other conversations with Matt have resulted in Allman braces everywhere (probably)
Classes & Namespaces: CamelCase
Member variables: snake_case_with_trailing_underscore_
Member methods: snake_case()
Variables: snake_case

zdroid:
@ozmorph There's not really much to argue about there. Diff logs look horrible with tabs, and especially with tabs and spaces mixed. It is configurable, but most of the users will never bother to do that. In case they see a horrible looking code, they won't keep looking.

Using spaces instead of tabs allows a more general config, like `.editorconfig`, to be used for all textual files (R, Python, C++, txt). All coding styles supported by VS Code (Google, LLVM, Visual Studio, Chromium, Mozilla, WebKit) use spaces instead of tabs.

CloneDeath:
> During a previous conversation with @matt-gretton-dann he said he'd put together a clangformat file. Other conversations with Matt have resulted in Allman braces everywhere (probably)
> Classes & Namespaces: CamelCase
> Member variables: snake_case_with_trailing_underscore_
> Member methods: snake_case()
> Variables: snake_case

[...]

PR #307, opened 2020-05-20
Refactor the Kernel

zebmason:
Refactor the code in Kernels.h & .cpp to C++ code

### DIFF
--- a/src/Kernels.cpp
+++ b/src/Kernels.cpp
@@ -3,68 +3,70 @@
 #include "Kernels.h"
 #include "Error.h"
 #include "Dist.h"
-#include "Param.h"
 
-#include "Model.h"
+using namespace CovidSim::TBD1;
 
-// To speed up calculation of kernel values we provide a couple of lookup
-// tables.
-//
-// nKernel is a P.NKR+1 element table of lookups nKernel[0] is the kernel
-// value at a distance of 0, and nKernel[P.NKR] is the kernel value at the
-// largest possible distance (diagonal across the bounding box).
-//
-// nKernelHR is a higher-resolution table of lookups, also of P.NKR+1
-// elements.  nKernelHR[n * P.NK_HR] corresponds to nKernel[n] for
-// n in [0, P.NKR/P.NK_HR]
-//
-// Graphically:
-//
-// Distance 0            ...                              Bound Box diagonal
-//          nKernel[0]   ... nKernel[P.NKR / P.NK_HR] ... nKernel[P.NKR]
-//          nKernelHR[0] ... nKernelHR[P.NKR]
-double *nKernel, *nKernelHR;
-
-void InitKernel(double norm)
+bool KernelLookup::setup(double longest_distance)
 {
-	double(*Kernel)(double);
-
-	if (P.KernelType == 1)
-		Kernel = ExpKernel;
-	else if (P.KernelType == 2)
-		Kernel = PowerKernel;
-	else if (P.KernelType == 3)
-		Kernel = GaussianKernel;
-	else if (P.KernelType == 4)
-		Kernel = StepKernel;
-	else if (P.KernelType == 5)
-		Kernel = PowerKernelB;
-	else if (P.KernelType == 6)
-		Kernel = PowerKernelUS;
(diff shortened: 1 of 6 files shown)


zebmason:
This is just a heads up so I don't find anyone else doing the same refactoring

CloneDeath:
Ha, I was actually in the middle of a VERY similar change, before backing out to work on something else.

You have my approval, I look forward to seeing it when done :)

zebmason:
@matt-gretton-dann Are the naming conventions for structs the same as for classes?

CloneDeath:
@matt-gretton-dann Yes, PascalCase for structs as well as classes. And make sure to use snake_case for class/struct functions.

zebmason:
And snake_case_ for public member variables?

CloneDeath:
snake_case_ (with underscore) for private,

I **think** @matt-gretton-dann wanted just snake_case (no underscore) for public, but I can't seem to find where I saw that...

Honestly, I wouldn't stress *too* much about it. If Matt has an issue with something you've done, just make sure to note it in #305 so we can make sure to include it in our style guide.

We're all trying to figure out how things ought to be as a community, you're definitely not alone in this :)

zebmason:
@dlaydon Could you look at the comments in Kernels.h or suggest someone else, please?

src/Kernels.cpp
```diff
@@ -3,132 +3,136 @@
...
-		Kernel = GaussianKernel;
-	else if (P.KernelType == 4)
-		Kernel = StepKernel;
-	else if (P.KernelType == 5)
-		Kernel = PowerKernelB;
-	else if (P.KernelType == 6)
-		Kernel = PowerKernelUS;
-	else if (P.KernelType == 7)
-		Kernel = PowerExpKernel;
+	try
```
CloneDeath, line 11: I think this is a case of tell don't ask.

Let the function throw an exception instead of returning a boolean.

zebmason: Did think about that at the time but I'm always loath to change the behaviour of unfamiliar code so I thought I'd do the ugly solution. I'll change it and not do any check on return. Thanks.

zebmason:
@weshinsley Ready for review

src/Kernels.h
```diff
@@ -1,16 +1,122 @@
...
+      /// \return Probability
+      double step(double r2) const;
+    };
+
+    /// \brief To speed up calculation of kernel values we provide a couple of lookup
+    /// tables.
+    ///
+    /// lookup_ is a table of lookups. lookup_[0] is the kernel
+    /// value at a distance of 0, and lookup_.back() is the kernel value at the
+    /// largest possible distance (diagonal across the bounding box).
```
matt-gretton-dann, line 67: `(diagonal across the bounding box)` is incorrect with spherical co-ordinates.  See https://github.com/mrc-ide/covid-sim/blob/master/src/SetupModel.cpp#L201

zebmason: Bit obvious I suppose - the bounding box could straddle the equator and we do live on an oblate spheroid.

zebmason:
@matt-gretton-dann I did think this would just be a first commit to move the code around and thus be functionally identical. I had intended to re-do the numerics of the table calculation as they waste a lot of flops and I didn't see the compiler ever being intelligent enough to optimise them but I suppose they're not important if the table only needs calculating the once.

weshinsley:
I don't know why the CI for macOS is doing this - no-one has cancelled it...

zebmason:
@weshinsley It used to be that one of the Linux containers used to time out. Rather a pain but I usually just amend the last commit to create a new hash and then force push. Have a good weekend.

zebmason:
Unit tests look a tough one. The actual kernel functions can be tested with a few different values but they are trivial apart from discerning appropriate values. I'd kick it back to Imperial and ask them to think on the tests.

zebmason:
@weshinsley, @zdroid Interestingly this PR had already implemented all the changes in #419 some time ago

weshinsley:
Yeah, I know... PR #419 was a fairly easy one for me to review...! This one may be better for Matt to finish off...

PR #310, opened 2020-05-20
Move microcell math into functions

CloneDeath:
I'm working on refactoring assigning people to places, when I noticed that this microcell position code could be refactored out.

I think there's some more we could do in general, but I think this is a good first step in the right direction.


matt-gretton-dann:
@CloneDeath: Can you rename the .hpp files to .h to be consistent with the other header files please?

PR #312, opened 2020-05-21
Factor out domain sizes

zebmason:


dlaydon:
@zebmason really like this, approved!

zebmason:
@dlaydon Have you tried out the new doxygen support yet?
Type "make doxygen" in Linux or set the BUILD_DOC CMake setting to ON in Visual Studio.
Start to see the class list expanding

PR #313, opened 2020-05-21
Add macOS to the OS list

zdroid:


README.md
```diff
@@ -36,7 +36,7 @@ to:
 
 ## Building
 
-The model is written in C++ and runs on Windows and Linux-based systems.
+The model is written in C++ and runs on Windows, macOS and Linux-based systems.
```
matt-gretton-dann, line 39:
```suggestion
The model is written in C++.

The primary platforms it has been developed and tested on are Windows and Ubuntu Linux.

It should build and run on any other POSIX compliant Unix-like platform (for example macOS, other Linux distributions).  However, no active development occurs on them.
```

I want to set expectations about where we have tested.  Yes the CI loop tests on more platforms than Ubuntu Linux & Windows, but that is not all the testing that occurs.

zdroid:
@matt-gretton-dann Done.

weshinsley:
Merging as it's a README change that everyone's happy with...

PR #318, opened 2020-05-21
Factor out initial infections

zebmason:


src/Param.h
```diff
@@ -22,6 +22,20 @@ struct DomainSize
 	double height_;
 };
 
+struct InitialInfections
```
CloneDeath, line 25: I think this should be put into a different file.

src/Param.h
```diff
@@ -22,6 +22,20 @@ struct DomainSize
...
+{
+  int number_[MAX_NUM_SEED_LOCATIONS];
+  int do_random_loc_;
```
CloneDeath, line 28: I wouldn't abbreviate. `do_random_location`

src/Param.h
```diff
@@ -22,6 +22,20 @@ struct DomainSize
...
+  int do_all_in_same_loc_;
+  int min_pop_dens_;
```
CloneDeath, line 30: Same as above, don't abbreviate. `min_population_density`

I'm okay with using abbreviations for things that have a smaller scope, but for member variables that are used throughout, I prefer to make it clearer to the reader.

CloneDeath: I'd maybe even go so far as even expand `min` to `minimum`...

But if you disagree, I'd at LEAST like it to be `min_pop_density` :)

zebmason:
@matt-gretton-dann Just initial grouping of logically related values so as to reduce the number of elements in params.

zebmason:
@matt-gretton-dann Perhaps I should have mentioned that I've been further developing my code analysis tool to factor out classes from dependencies so my commits since I stopped playing with GoogleTest (I've made it safe for CI by commenting out) have mainly been the result of testing the new tool.

src/Param.h
zebmason: So do I but only when the code base is more modular

src/Param.h
zebmason: Done

src/Param.h
zebmason: Done

Issue #309, opened 2020-05-20
Potential Data Races Detected by Static Code Scanner


BradSwain:
Hello, we are working on a race detection tool and found two potential data races in this project.
The races we found do not appear to be dangerous to the core covid-sim model, but we thought it would be better to report them just in case as data races could potentially lead to undefined behavior, especially at high optimization levels.

## Race on `nevents`
The first race is on `nevents`. There is a write guarded by a critical section at [Update.cpp:215](https://github.com/mrc-ide/covid-sim/blob/master/src/Update.cpp#L215) and an unguarded read at [Update.cpp:145](https://github.com/mrc-ide/covid-sim/blob/master/src/Update.cpp#L145).
```
==== Found a race between: 
line 215, column 3 in src/Update.cpp AND line 145, column 8 in src/Update.cpp
Shared variable: 
 at line 287 of src/SetupModel.cpp
 287|		if (!(nEvents = (int*)calloc(1, sizeof(int)))) ERR_CRITICAL("Unable to allocate events storage\n");
Thread 1: 
 213|
 214|		//increment the index of the infection event
>215|		(*nEvents)++;
 216|	}
 217|
>>>Stack Trace:
>>>DoInfect(int, double, int, int) [src/Sweep.cpp:717]
>>>  RecordEvent(double, int, int, int, int) [src/Update.cpp:147]
Thread 2: 
 143|		if (P.DoRecordInfEvents)
 144|		{
>145|			if (*nEvents < P.MaxInfEvents)
 146|			{
 147|				RecordEvent(t, ai, run, 0, tn); //added int as argument to RecordEvent to record run number: ggilani - 15/10/14
>>>Stack Trace:
>>>DoInfect(int, double, int, int) [src/Sweep.cpp:717]
The OpenMP region this bug occurs:
/git/covid-sim/src/Sweep.cpp
>701|#pragma omp parallel for schedule(static,1) default(none) \
 702|		shared(t, run, P, StateT, Hosts, ts)
 703|	for (int j = 0; j < P.NumThreads; j++)
 704|	{
 705|		for (int k = 0; k < P.NumThreads; k++)
 706|		{
Gets called from:
>>>main
>>>  RunModel(int) [src/CovidSim.cpp:409]
```
It seems this race can only occur if `P.DoRecordInfEvents` is set, and the most likely negative effect looks to be a potential overflow of `InfEventLog` log.

## Race on `State.cumV`
The second race we found is on `State.cumV`.

[...]

PR #320, opened 2020-05-21
Read/write State.cumV safely

igfoo:
All the cumV reads are now also in critical sections, thus fixing one of the races identified in #309.

I don't know if the overhead of doing the `cumV` check in `DoDetectedCase` is less than work than is saved; it'll depend on how often the check succeeds. But I thought it best to leave the code in, as it's easier to remove it in the future than to re-add it.

### DIFF
--- a/src/Update.cpp
+++ b/src/Update.cpp
@@ -1259,16 +1259,20 @@ void DoPlaceOpen(int i, int j, unsigned short int ts, int tn)
 void DoVacc(int ai, unsigned short int ts)
 {
 	int x, y;
+	bool cumV_OK = false;
 
 	if ((HOST_TO_BE_VACCED(ai)) || (Hosts[ai].inf < InfStat_InfectiousAlmostSymptomatic) || (Hosts[ai].inf >= InfStat_Dead_WasAsymp))
 		return;
-	if (State.cumV >= P.VaccMaxCourses)
-		return;
+#pragma omp critical (state_cumV)
+	if (State.cumV < P.VaccMaxCourses)
+	{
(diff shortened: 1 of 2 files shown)


NeilFerguson:
Thanks - yes best to include

Best,

Neil
________________________________

src/Update.cpp
```diff
@@ -610,12 +610,20 @@ void DoDetectedCase(int ai, double t, unsigned short int ts, int tn)
...
-		if ((!P.DoMassVacc) && (t >= P.VaccTimeStart) && (State.cumV < P.VaccMaxCourses))
-			if ((t < P.VaccTimeStart + P.VaccHouseholdsDuration) && ((P.VaccPropCaseHouseholds == 1) || (ranf_mt(tn) < P.VaccPropCaseHouseholds)))
+		if ((!P.DoMassVacc) && (t >= P.VaccTimeStart))
+		{
+			// DoVacc is going to test that `State.cumV < P.VaccMaxCourses` itself before
+			// incrementing State.cumV, but by checking here too we can avoid a lot of
+			// wasted effort
+			bool cumV_OK;
+#pragma omp critical (state_cumV)
+			cumV_OK = State.cumV < P.VaccMaxCourses;
```
matt-gretton-dann, line 620: It is not obvious to me whether this critical section lasts for the rest of the block or just the next statement.  I believe we want just the statement - so can we put it in a block just to make it obvious please?

weshinsley:
Looks good to me, and this would be good to have in place. Also satisfies Matt's only comment...

PR #321, opened 2020-05-21
Test State.cumVG inside the critical section

igfoo:
This removes the race, mentioned at the end of #309, where another thread might increment it between our test and use of it.

### DIFF
--- a/src/Update.cpp
+++ b/src/Update.cpp
@@ -1300,7 +1300,7 @@ void DoVaccNoDelay(int ai, unsigned short int ts)
 	int x, y;
 	bool cumVG_OK = false;
 
-	if ((!HOST_TO_BE_VACCED(ai)) && (Hosts[ai].inf >= InfStat_InfectiousAlmostSymptomatic) && (Hosts[ai].inf < InfStat_Dead_WasAsymp))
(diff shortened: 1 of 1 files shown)


src/Update.cpp
```diff
@@ -1298,12 +1298,19 @@ int DoVacc(int ai, unsigned short int ts)
 void DoVaccNoDelay(int ai, unsigned short int ts)
 {
 	int x, y;
+	bool cumVG_OK = false;
 
-	if ((State.cumVG < P.VaccMaxCourses) && (!HOST_TO_BE_VACCED(ai)) && (Hosts[ai].inf >= InfStat_InfectiousAlmostSymptomatic) && (Hosts[ai].inf < InfStat_Dead_WasAsymp))
+	if ((!HOST_TO_BE_VACCED(ai)) && (Hosts[ai].inf >= InfStat_InfectiousAlmostSymptomatic) && (Hosts[ai].inf < InfStat_Dead_WasAsymp))
```
weshinsley, line 1303: Should the conditional on 1303 to be wrapped in a NOT - so that 1304 exits the function if one or more of the inner conditionals fails?

igfoo:
Good spot, thanks! I've negated all the logic, so we don't have a negation of a negation. This also makes it now match the check in `DoVacc`.

src/Update.cpp
weshinsley: Fix looks great :-)

PR #325, opened 2020-05-22
Create Vector2 Struct

CloneDeath:
Refactor a lot of xy logic into some basic structures.
This includes a Vector2 and Size struct to hold x,y coordinates, as well as width/height.


covid-sim.vcxproj
```diff
@@ -123,8 +128,13 @@
...
     <ClInclude Include="src\Sweep.h" />
     <ClInclude Include="src\Update.h" />
-    <ClInclude Include="src\MicroCellPosition.hpp" />
     <ClInclude Include="src\Direction.hpp" />
+    <ClInclude Include="src\Household.hpp" />
+    <ClInclude Include="src\Cell.hpp" />
+    <ClInclude Include="src\Coordinates\MicroCellPosition.hpp" />
+    <ClInclude Include="src\Coordinates\Location.hpp" />
+    <ClInclude Include="src\Coordinates\Vector2.hpp" />
+    <ClInclude Include="src\Coordinates\Size.hpp" />
```
matt-gretton-dann, line 137: .h please.

src/Coordinates/Location.hpp
```diff
@@ -0,0 +1,8 @@
+#pragma once
+
+#include "Vector2.hpp"
+
+struct Location : Vector2<float> {
+	Location(float x, float y) : Vector2<float>(x, y){};
+};
+
```
matt-gretton-dann, line 8:
```suggestion
using Location = Vector2<float>;
```

src/Household.hpp
```diff
@@ -0,0 +1,14 @@
...
+
+struct Household
+{
+	int FirstPerson;
+	unsigned short int nh; // number people in household
+	Location loc;
+	unsigned short int nhr;
+
+	float distance_to(Household other);
+	float distance_squared_to(Household other);
```
matt-gretton-dann, line 13:
```suggestion
	float distance_to_squared(Household other);
```

matt-gretton-dann: Its the square of "distance_to"

zebmason:
See #213

src/Coordinates/Vector2.cpp
```diff
@@ -0,0 +1,39 @@
...
+	return Vector2<T>(std::abs(this->x), std::abs(this->y));
+}
+
+template<class T>
+Vector2<T> Vector2<T>::operator+(Vector2<T> other) const {
+	return Vector2<T>(this->x + other.x, this->y + other.y);
+}
+
+template<class T>
+Vector2<T> Vector2<T>::operator-(Vector2<T> other) const {
```
Venetian, line 41: nit: I think the operators should be passed vector by const ref, i.e. `const Vector2<T>& other` to prevent them making a copy. https://en.cppreference.com/w/cpp/language/operators

zebmason: I think Eigen should be used as it does delayed evaluation otherwise we are looking at a hit in speed hence why I coded up #213

covid-sim.vcxproj
CloneDeath: Fixed. hpp renamed to h

src/Coordinates/Location.hpp
CloneDeath: This ended up going away as I fleshed out Vector2 and Size (commits before this had operator overloads in Location, before I created Vector2 base class, and pulled the functions into there).

This commit must have been while I was still working on that.

src/Household.hpp
CloneDeath: Yeah, that makes sense when you put it like that...

In my head, I was thinking it's the "distance squared"... to a Household. But no, "distance to a Household, squared" makes MUCH more sense and flows way better.

Fixing this now.

CloneDeath:
> I strongly encourage the use of namespaces as appropriate.

I placed namespaces around the Geometry and Model classes. I don't think it's great, but it's a start.

I'd like to dive deeper into better namespaces and more organization later, but for now, I think this is a good first step.

CloneDeath:
@matt-gretton-dann 
> I also don't like lots of little header files.

I think the files in Geometry should be kept apart, as their differences are significant enough, and I wouldn't want people to confuse (for example) a function for Vector2 with Size.

As for the Models, I believe that as we start refactoring more and more functions into these classes, that their disparity will grow. However, as it stands, a few are still very small and similar. So, if you think we should merge some of these classes, let me know, and I'll get on it right away.

CloneDeath:
@zebmason 
> See #213

This avoids the issue of bringing in a third party library, but I did like your name for the Geometry objects.

This PR has gotten very large, but I have skimmed through that PR and saw a few things I'd like to pull over too.

Is there anything specifically you think should go in this PR, and not wait for a followup?

src/Coordinates/Vector2.cpp
CloneDeath: 1) I believed the slowdown (when I was writing that) to be negligable, a Vector2 is just two doubles.

2) Despite that, I switched it to a const ref later on. I don't have any performance metrics on that, though.

zebmason:
@CloneDeath I found with #213 that it just started to grow and grow then the merge conflicts turned it to hell. I like the bounding box code that I wrote as it actually flagged up and inconsistency in the code #268

PR #326, opened 2020-05-22
Better max value on P.latent_icdf

weshinsley:
Remove a warning...


dlaydon:
Looks good, but will wait for tests to pass to guard to be safe

weshinsley:
Actually, this is not the right answer - the parameter should be a double, and that is already merged in, so 1e10 is fine. (Some DOUBLE_MAX might be nicer, but there's no functional issue here...)

PR #329, opened 2020-05-23
Create paramaters-description.md

derekwinton:
This file explains the purpose of the paramaters class, groups variables by type to aid legibility and provides a description of each variable

### DIFF
--- a/docs/paramaters-description.md
+++ b/docs/paramaters-description.md
@@ -1,0 +1,184 @@
+# Paramaters
+
+WIP - interpretations below are based on reading the SetupModel.cpp file
+
+The paramaters class contains:
+- configurable options for running the simulation 
+- quantitative information used in the simulation
+- data produced about the setup process e.g. running totals
+- variables used to load/save information
+- variable used to manage processor / memory
+
+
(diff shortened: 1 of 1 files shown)


insidedctm:
This seems like a good idea. You might want to keep an eye on some of the int parameters changing to bool- I can seem some changes in the param.h overnight.

Also parameter not paramater.

derekwinton:
Good catch on spelling,  I have updated thanks. 

I wasn't sure how to handle the bool / int distinction. They are integer values but treated as booleans ie 0= Off, 1=0n.

insidedctm:
I’m guessing there is a general desire to change variables to bool if that is how they are used. Makes sense.

weshinsley:
A few pre-review thoughts on this 
(1) Important to get started on this - thanks @derekwinton for kicking it off (some time ago...)
(2) I know a bunch of work has been going on documenting this, which will be on its way into a github form and may supercede this PR.
(3) I'd be excited about any creative way that the code could be self-documenting - possibly some pipline that updates this .md and keeps it in sync, as I imagine it easy to let these things slide.

PR #340, opened 2020-05-26
Add comments to InfectSweep algorithm

derekwinton:
The InfectSweep function performs infections in: Households, Places, Cells

I added comments to the Households section to identify and explain the variables used and also to identify loop and conditional closing braces


src/Sweep.cpp
```diff
@@ -282,7 +282,7 @@ void InfectSweep(double t, int run) //added run number as argument in order to r
 	int f, f2, cq /*cell queue*/, bm, ci /*person index*/;
 	double seasonality, sbeta, hbeta;
 	//// various quantities of force of infection, including "infectiousness" and "susceptibility" components
-	double s; // household FOI on fellow household member, then place susceptibility, then random number for spatial infections allocation* / ;
+	double s; // household Force Of Infection (FOI) on fellow household member, then place susceptibility, then random number for spatial infections allocation* / ;
```
CloneDeath, line 285: I think the variable should be split into three variables, instead of re-used. 
Then, each variable should be named appropriately.

src/Sweep.cpp
```diff
@@ -291,131 +291,246 @@ void InfectSweep(double t, int run) //added run number as argument in order to r
 	double fp; //// false positive
 	unsigned short int ts;
 
-	if (!P.DoSeasonality)	seasonality = 1.0;
-	else					seasonality = P.Seasonality[((int)t) % DAYS_PER_YEAR];
+	// if not doing seasonality
+	if (!P.DoSeasonality)
+	{	
+		// set seasonality to 1
```
CloneDeath, line 297: I don't think this sort of comment (where the line of code is replicated in english) adds anything useful.

src/Sweep.cpp
```diff
@@ -291,131 +291,246 @@ void InfectSweep(double t, int run) //added run number as argument in order to r
...
+		seasonality = 1.0;
+	}
+	else{
+		// otherwise pick seasonality from P.Seasonality array using day number in year
```
CloneDeath, line 301: I'd like to see this refactored into a function with that sort of name (ie `pick_seasonality_from_day_number(int day_number)`), rather than described in a comment.

src/Sweep.cpp
```diff
@@ -291,131 +291,246 @@ void InfectSweep(double t, int run) //added run number as argument in order to r
...
+		seasonality = P.Seasonality[((int)t) % DAYS_PER_YEAR];
+	}
 
-	ts = (unsigned short int) (P.TimeStepsPerDay * t);
+	ts = (unsigned short int) (P.TimeStepsPerDay * t); // the timestep number of the start of the current day
+	// fp = false positive
```
CloneDeath, line 306: Another instance where I'd rather see the variable renamed as `false_positive` instead of a comment.

src/Sweep.cpp
derekwinton: I don't disagree. My initial approach was not to make any code changes but just add comments to aid understanding of the code 'as is' with the hope that this would aid refactoring.

CloneDeath: I figured something like that. In that case, I approve. :)

src/Sweep.cpp
derekwinton: In an ideal situation I'd agree with you. My reasoning for doing so was that the variable names don't always reflect what they represent so I felt there was value in spelling out explicitly what was occurring. I'm not wedded to the idea all that strongly though.

src/Sweep.cpp
derekwinton: Absolutely agree. There is ample scope for refactoring.

weshinsley:
No behaviour changes. just comment-only, and also a couple of bracket-expansions to make style/spacing more consistent. All commenting is probably WIP, along with other refactors, so happy to merge as a small step forward...

PR #344, opened 2020-05-28
Redesign of InfStat into a state design pattern

mstevens-uk:
This is a DRAFT Proof of Concept of redesigning InfStat into a state pattern design.

Each InfStat enum gets its own class.
Each infection state can either get better or get worse. (This is used to model state transitions.)

This would be the first step towards redesign.
The intention would be to restrict this change to the state pattern only.
There would be no changes to the underlying logic.


mstevens-uk:
Closing this Draft PR.
Lessons:

- it would be very difficult to review changes where significant blocks of code are moved into new classes, with the originals continually being updated 

- The proposed design lost context of each state transition without significant benefit, as the underlying code remained in Update.cpp and Sweep.cpp.

PR #316, opened 2020-05-21
Converted Param.h icdf arrays from double array to InverseCdf class

mstevens-uk:
Closes #315. (Refactor the ICDF arrays in Param.h / CovidSim.cpp)

- Introduces a new class InverseCdf
- Removes SetICDF() from CovidSim.cpp
- Adds helper function GetInverseCdf() to replace individual calls
- Uses InverseCdf functions set_defaults() and apply_exponent()
- Moves ChooseFromICDF() from Update.cpp to InverseCdf function choose()

Issue #74, opened 2020-04-09
Make parameter handling more robust


matt-gretton-dann:
Command line option handling and reading of input files do not give helpful errors when inputs are incorrect.  This issue is to work through some of these and improve command-line and input file parsing.  Ideally adding unit/regression tests as we go:

 * [ ] Ensure for command line options taking an input file name we check the file exists
 * [ ] Make parameter file handling robust:
    * [ ] When using values passed via /CLP#: erroring if /CLP# wasn't specified on command line
    * [ ] Issuing warnings about what options weren't used.
    * [ ] Issuing warning if we can't parse the value for a parameter
    * [ ] Highlighting where one value overrides another
    * [ ] Enable output of a parameter file which merges all values read.

Feynstein:
I've also noticed that some parameters can be subsets of other parameters in order to shrink the size of the param strict. The image was used for another issue but you can see some of those parameters that could be aggregated.
![81463760-072b1600-918a-11ea-817f-93012f1887af](https://user-images.githubusercontent.com/29045335/81464444-dea61a80-918f-11ea-8fac-da6fbdd2b0b8.png)

PR #346, opened 2020-05-28
More robust parameter file handling

ozmorph:
Part 2 of #74. This pull is related to my other PR #228 which deals with the CLI parameter parsing aspects of #74. You may note that the `raw_extract()` number extraction code from `ParamFile.hpp` looks similar to the `parse_number()` function in PR #228. Eventually these two implementations could be combined into a separate file, but I think they can exist independently for now.

The main goals of this Pull-Request are:

- A more clean and concise way of extracting parameter values (fewer 160+ character lines)
- A type-safe approach towards extracting values
  - Logging when a value overflows/underflows the storage size of the output variable
  - Replacing the use of `fscanf()` and `sprintf()` with `std::ifstream` and `std::istringstream`.
- Logging when the actual number of parameters extracted doesn't match the expected
- Logging when a parameter value isn't specified and a default value is instead used
- Exiting when a required parameter value isn't found

Things that are missing from this draft PR right now (I will update as I make more commits):

- [ ]  Wildcard CLI parameters (e.g. specifying `#1` in a param file and using `/CLP1` value from the CLI)
- [ ]  Extracting string values
- [ ]  Extracting matrix values (values that span more than 1 line)
- [ ]  Extracting values into a proportion array where the default value is `1` at index 0 and `0` in every other place.
- [ ]  Extracting values into an array with a default array specified (e.g. `output[4]` can have `default{ 1, 2, 3, 4}`)
- [ ]  Extracting ICDF values. I'm currently watching #316 
- [ ]  Outputting a parameter file with all merged and de-duplicated values

I will continue to work on this PR, but please let me know if there are any design decisions you would like to discuss or changes you would like to be made!

Cheers!


ozmorph:
It appears that old ABI of the libstdc++ used in the Docker CentOS 7 container doesn't allow for the C++11 function `std::string::erase(const_iterator)` to be used, which is why that CI check is failing.

Either `-DGLIBCXX_USE_CXX11_ABI=1` needs to be passed to the command-line for that build, or an alternative should be found and used.

References:

[GCC Bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=81967)
[StackOverflow](https://stackoverflow.com/questions/47912089/why-is-the-wrong-gcc-7-2-libstdc-abi-being-selected-on-centos-7)

src/ParamFile.cpp
```diff
@@ -0,0 +1,273 @@
...
+#include <sstream>
+#include <string>
+
+#include "ParamFile.hpp"
+
+// ltrim(), rtrim(), and trim() are from a highly up voted answer on StackOverflow
+// https://stackoverflow.com/a/217605
+
+// trim from start (in place)
+static inline void ltrim(std::string& s) {
```
matt-gretton-dann, line 17: No need for `inline` here - its only a hint to the compiler - which will ignore it.

matt-gretton-dann:
Things I would like to see in this PR:

 * Detection and warning of when we do have multiple values for a field
 * Detection and warning of when a parameter is specified in a config file but not read.
 * A new command-line option that enables output of a parameter file that when used as the input will result in the same config.

PR #348, opened 2020-05-29
Properly scope an if in Bitmap_Finalise

zdroid:


weshinsley:
Fine - simple change...

PR #357, opened 2020-06-01
Minor tidying

NeilFerguson:
Removing spurious tabs and new lines


weshinsley:
Also - can I double-check the changes on CovidSim.cpp:5007, 5012, 5027 and 5046 are intentional? They look like more than a tidying change, and I'm wondering whether they are intended for this PR.

NeilFerguson:
Yes, slightly more. But would be good to have them.

Best,

Neil
________________________________

src/CovidSim.cpp
```diff
@@ -5004,12 +4954,12 @@ void RecordSample(double t, int n)
 				{
 					s = ((double)trigAlert)/((double)P.AlertTriggerAfterIntervThreshold);
 					thr = 1.1 / sqrt((double)P.AlertTriggerAfterIntervThreshold);
-					if (thr < 0.05) thr = 0.05;
+					if (thr < 0.05) thr = 0.04;
```
matt-gretton-dann, line 4957: Is this change correct or incomplete - it looks odd?  Did you mean to change the 0.05 to 0.04 - or did you miss the condition check earlier and fail to change that too?

Issue #358, opened 2020-06-01
Lack of determinism in regression test results when compiled with the Intel compiler on Linux


owainkenwayucl:
*(I'm digging into this so hope to have more detail later)*

Currently, contradictory to what is said in #161, and in [report9/README.md](https://github.com/mrc-ide/covid-sim/tree/master/report9), the codebase is not deterministic when compiled with the Intel compiler suite (tested with version 2020 (release) and 2019 (update 5) on RHEL 7.8).

> "Covidsim is now deterministic across platforms (linux, Mac and Windows) and across compilers (gcc, clang, intel and msvc) for a specified number of threads and fixed random number seeds."

Releases up to and including 0.8.0 could be built and pass the regression test suite on newer Intel compilers such as those provided with the Intel 2019 Update 5 and 2020 (release) compiler suites but newer versions of CovidSim no longer pass (they are fine with the GNU compiler suite, at a significant performance penalty so we deploy GNU builds on our HTC/HPC resources).

Out of mild interest, and based on the above statement I figured I'd try again with the Intel compilers and now not only do checksums not match (presumably due to minor floating point differences), **but the code (both 0.14.0 and master) reliably writes out different numbers of files in the  [UK regression test suite script](https://github.com/mrc-ide/covid-sim/blob/master/tests/regressiontest_UK_100th.py)** when compiled with either the 2020 or 2019 compiler suite.  This produces the error:

```
Lengths don't match.
```

in the test output.  Sometimes, the test writes out more files than are in the reference set but sometimes it writes fewer.  This looks like it's related to run length (e.g. when it's shorter it's, for example, the end files of `mt-*.bmp` etc that are missing) so presumably(?) this is the simulation taking shorter or longer to converge due to accumulated floating point errors.

This seems somewhat dependent on the micro architecture - for example 0.14.0 writes more files on one of our compute nodes (Cascade Lake) but fewer on one of the login nodes (Skylake).

_Edit_ - *On further testing this difference between the two platforms seems to be related to `$OMP_NUM_THREADS` being set explicitly on the compute nodes inside a job whereas it's not set on the login nodes, however `CovidSim` is told to use 1 thread, so I guess the compiler is doing ... something.*

This behaviour occurs regardless of whether I run it directly (have it build its own binary) or run the tests by running `make test` in a build directory.  It does not seem to happen with the US test, which merely has the predicted checksum issues.

[...]

PR #360, opened 2020-06-02
Set -fp-model precise when using the ICPC+Linux

owainkenwayucl:
As per discussion in https://github.com/mrc-ide/covid-sim/issues/358 this adds the flag `-fp-model precise` to the C++ compiler flags when compiling with the Intel compiler on Linux which aids passing regression tests on that platform.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,5 +1,10 @@
 # CMakeLists.txt for src directory
 
+# Add -fp-model precise in to Intel compiler targets (must be done before target is defined).
+if(UNIX AND (CMAKE_CXX_COMPILER_ID STREQUAL "Intel"))
+  add_compile_options("-fp-model" "precise")
+endif()
+
 # Set up the IDE
 set(MAIN_SRC_FILES CovidSim.cpp BinIO.cpp Rand.cpp Error.cpp Dist.cpp Kernels.cpp Bitmap.cpp SetupModel.cpp CalcInfSusc.cpp Sweep.cpp Update.cpp Param.cpp MicroCellPosition.cpp Direction.cpp)
 set(MAIN_HDR_FILES CovidSim.h BinIO.h Rand.h Constants.h Country.h MachineDefines.h Error.h Dist.h Kernels.h Bitmap.h Model.h Param.h SetupModel.h ModelMacros.h InfStat.h CalcInfSusc.h Sweep.h Update.h MicroCellPosition.hpp Direction.hpp)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,10 +1,5 @@
 # CMakeLists.txt for src directory
 
-# Add -fp-model precise in to Intel compiler targets (must be done before target is defined).
-if(UNIX AND (CMAKE_CXX_COMPILER_ID STREQUAL "Intel"))
-  add_compile_options("-fp-model" "precise")
-endif()
-
 # Set up the IDE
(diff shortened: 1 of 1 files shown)


weshinsley:
I think this is a good idea - and possibly should be applied to the Windows Intel compiler as well. Hard to test without having an Intel C++ compiler for linux...

owainkenwayucl:
@weshinsley Similarly, the reason I restricted this patch to Linux is that I can't test it on Windows (and the Intel compiler options have a different format on Windows so I really would want to test it).

I have tested this on our Linux systems at it works as intended, i.e. on the Intel compilers it sets `-fp-model precise` and on the GNU + clang ones it doesn't.

owainkenwayucl:
It does *seem* to be trying to do something similar on lines 19-22 for Windows but in my reading of the documentation I'm not sure `add_compile_options("fp:strict")`actually does anything to targets declared before it is set on some versions of CMake, e.g. for 3.13.x:

> Adds options to the compiler command line for targets in the current directory and below that are added after this command is invoked. See documentation of the directory and target COMPILE_OPTIONS properties.

But then for 3.17.x it says:

> Adds options to the COMPILE_OPTIONS directory property. These options are used when compiling targets from the current directory and below.

so I'm guessing the behaviour has changed.

In the root `CMakeLists.txt` for the project the minimum version is 3.8, so I guess the Windows rules may or may not be applied correctly depending on what version people are using.

weshinsley:
I'll have to ask about this a bit more as far as cmakelists goes...

What I did find out though, is that when compiling with Intel from within the Visual Studio integration, VS's default (/fp:precise) is sent on the Intel compiler commandline by default. (Which is nice)

matt-gretton-dann:
There is `target_compile_options()` which I think should be also used for the Visual Studio FP precise statement lower down?

owainkenwayucl:
@weshinsley what should I do to progress this getting merged?

weshinsley:
Sorry @owainkenwayucl I'd forgotten about this - I'm not very familiar with cmake and I'd put it to one side to read up on.

I think what @matt-gretton-dann means is, move your add_compile_options statement to just after line 23 inside the elseif(UNIX), (and the `if` will get simpler)

Let's not't worry about Intel/Windows on this PR - I'll hook that up and test in a bit...

owainkenwayucl:
Ah right - my brain was also not working as I misunderstood `target_compile_options` as being somehow what was already being done but that is of course `add_compile_options`.  CMake be like that.

owainkenwayucl:
So I tried with `target_compile_options` on the Unix side and it works so it may be worth adapting the Windows part appropriately:

```diff
--- CMakeLists.txt.orig 2020-07-20 11:47:38.000000000 +0100
+++ CMakeLists.txt      2020-07-20 11:47:46.000000000 +0100
@@ -21,6 +21,9 @@
   endif()
 elseif(UNIX)
   target_compile_definitions(CovidSim PUBLIC UNIX)
+  if(CMAKE_CXX_COMPILER_ID STREQUAL "Intel")
+    target_compile_options(CovidSim PUBLIC "SHELL:-fp-model precise")
+  endif()
 else()
   message(FATAL_ERROR "Unknown operating system type: ${CMAKE_SYSTEM_NAME}")
 endif()
```

owainkenwayucl:
(Just to confirm - I've tested that this does the right thing both with Intel Compilers and GCC on Linux)

weshinsley:
Yep, and all tests pass. Ok, let's get this one in, and I'll back to the Intel command-line options for Windows

PR #373, opened 2020-06-06
Abstract periodic xy code

zdroid:
Abstracts repeated 5 lines of code into an appropriate double function.


weshinsley:
Generally I'd agree - but 0.5 is not overly magical - we just want half of the width or height in question. I'm not convinced that pulling that out as a const double would add much clarity.

PR #375, opened 2020-06-07
Buffer changes to host.absent_start|stop_time, and commit single-threaded

weshinsley:
Intel Inspector reported a data race on the middle two if statements Update.cpp:1229, 1230 on current master, in threaded `DoPlaceClose`)- data races on which thread gets to those particular lines first; no other lines involved.
```
if ((HOST_AGE_YEAR(l) >= P.CaseAbsentChildAgeCutoff) && (abs(Hosts[l].inf) != InfStat_Dead))
{
  if (Hosts[l].absent_start_time > t_start) Hosts[l].absent_start_time = t_start;
  if (Hosts[l].absent_stop_time < t_stop) Hosts[l].absent_stop_time = t_stop;
  StateT[tn].cumAPA++;
  f = 1;
}
```
The reason is a rare case where two schools close, and a child from each school belong to the same household, and each thread wants to update the absentee timings simultaneously. As the two instructions are effectively max/min, the results will not be affected by this change, and neither is master non-deterministic. Yet, it is probably helpful to remove the warning from Intel Inspector and thread this anyway. 

I have buffered these changes in a threaded-array, see `host_closure_queue` and `host_closure_queue_size` in Model.h, which are put in `StateT[tn]`. I then replace the above lines in `DoPlaceClosure` with thread-safe lines, and I call a new function `UpdateHostClosure()` to flush those changes, at the end of each surrounding thread-loop where `DoPlaceClosure` might be called. 

I believe since t_start and t_stop are a few timesteps in the future compared to the current timestep, the effects of the change do not need to be in place immediately (and we would see more warnings from Intel Inspector if that were necessary I think).

### DIFF
--- a/src/CovidSim.cpp
+++ b/src/CovidSim.cpp
@@ -2807,7 +2807,10 @@ void InitModel(int run) // passing run number so we can save run number in the i
 	{
 		for (int j = 0; j < MAX_NUM_THREADS; j++)	StateT[i].n_queue[j] = 0;
 		for (int j = 0; j < P.PlaceTypeNum; j++)	StateT[i].np_queue[j] = 0;
+		StateT[i].host_closure_queue_size = 0;
 	}
+
+	
 	if (DoInitUpdateProbs)
 	{
 		UpdateProbs(0);
@@ -2969,6 +2972,11 @@ void SeedInfection(double t, int* nsi, int rf, int run) //adding run number to p
 			}
 		}
 	}
+
+	// Is there a possibility that place closure can happen in the DoInfect calls above?
+	// Not sure - but to play safe...
+	UpdateHostClosure();
+
 	if (m > 0) fprintf(stderr, "### Seeding error ###\n");
 }
 
@@ -3079,6 +3087,7 @@ int RunModel(int run) //added run number as parameter
 									l = (int)(((double)P.PopSize) * ranf()); //// choose person l randomly from entire population. (but change l if while condition not satisfied?)
 								} while ((abs(Hosts[l].inf) == InfStat_Dead) || (ranf() > P.FalsePositiveAgeRate[HOST_AGE_GROUP(l)]));
 								DoFalseCase(l, t, ts, 0);
+								UpdateHostClosure();
 							}
 						}
 					}
--- a/src/CovidSim.cpp
+++ b/src/CovidSim.cpp
@@ -2973,10 +2973,6 @@ void SeedInfection(double t, int* nsi, int rf, int run) //adding run number to p
 		}
 	}
 
-	// Is there a possibility that place closure can happen in the DoInfect calls above?
-	// Not sure - but to play safe...
-	UpdateHostClosure();
-
 	if (m > 0) fprintf(stderr, "### Seeding error ###\n");
 }
 
@@ -3087,7 +3083,6 @@ int RunModel(int run) //added run number as parameter
 									l = (int)(((double)P.PopSize) * ranf()); //// choose person l randomly from entire population. (but change l if while condition not satisfied?)
 								} while ((abs(Hosts[l].inf) == InfStat_Dead) || (ranf() > P.FalsePositiveAgeRate[HOST_AGE_GROUP(l)]));
 								DoFalseCase(l, t, ts, 0);
-								UpdateHostClosure();
 							}
 						}
 					}
@@ -3106,7 +3101,9 @@ int RunModel(int run) //added run number as parameter
 						if ((!fs2) && (State.L + State.I == 0) && (P.FalsePositivePerCapitaIncidence == 0)) { if ((ir == 0) && (((int)t) > P.DurImportTimeProfile)) fs = 0; }
 					}
 					if (P.DoAirports) TravelReturnSweep(t);
+					UpdateHostClosure();
 				}
+
 				t += P.TimeStep;
 				if (P.DoDeath) P.ts_age++;
 				if ((P.DoSaveSnapshot) && (t <= P.SnapshotSaveTime) && (t + P.TimeStep > P.SnapshotSaveTime)) SaveSnapshot();
--- a/src/CovidSim.cpp
+++ b/src/CovidSim.cpp
@@ -2809,8 +2809,6 @@ void InitModel(int run) // passing run number so we can save run number in the i
 		for (int j = 0; j < P.PlaceTypeNum; j++)	StateT[i].np_queue[j] = 0;
 		StateT[i].host_closure_queue_size = 0;
 	}
-
-	
 	if (DoInitUpdateProbs)
 	{
 		UpdateProbs(0);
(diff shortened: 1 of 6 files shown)


weshinsley:
Tested in Intel Inspector with...
/c:8 /NR:10 /PP:data\TEST\preGB_R0=2.0new.txt /P:data\TEST\p_PC_CI_HQ_SD.txt /O:Output\PC_CI_HQ_SD_100_91_R0=2.2 /M:Output\TEST.bin /D:data\TEST\GB_pop2018.bin /L:data\TEST\NetworkGB_8T.bin /CLP1:100 /CLP2:91 /CLP3:121 /CLP4:121 /R:1.1 98798150 729101 17389101 4797132

and all clear in terms of threading issues. Checking no reproducibility breakages next...

weshinsley:
I have confirmed that all the report 9 runs produced with this branch are byte-identical to runs produced with the current master. 

This means two good things
a) This branch is looking good, and
b) The reported race in Intel Inspector never made any difference to results - as we already expected.

weshinsley:
@igfoo and @NeilFerguson - would especially appreciate you looking at my code changes - it's a fairly small change. 

Two particular things I coded "naively" -

1. I may be over-generous on SetupModel.cpp:1457 in the amount of space I am allocating to the buffer.

2. I may also be calling the flusher (UpdateHostClosure) in more places than I strictly need to; I suspect once per timestep might be sufficient, whereas I am calling at the end of each parallel section where changes can be made. However, calling UpdateHostClosure when the buffers are empty will take a negligible amount of time, so I went for "better safe than sorry".

src/Model.h
```diff
@@ -90,6 +90,17 @@ struct ContactEvent
 	unsigned short int contact_time;
 };
 
+/**
+ * @brief Apply place closure effects to household in a clearer way, to avoid *the appearance* of a data race.
+ *
+ */
```
igfoo, line 96: There is actually a real race that can happen: If you have one thread doing
```
if (absent_start_time > 5)
    absent_start_time = 5;
```
and another doing
```
if (absent_start_time > 7)
    absent_start_time = 7;
```
at the same time with `absent_start_time` initially 9 then these could run as
```
if (absent_start_time > 7)
if (absent_start_time > 5)
    absent_start_time = 5;
    absent_start_time = 7;
```
leaving it set to 7, rather than 5. So I'd suggest changing the comment to `Apply place closure effects to household in a thread-safe way.`

src/Model.h
```diff
@@ -110,6 +121,8 @@ struct PopVar
 	int cumItype[INFECT_TYPE_MASK], cumI_keyworker[2], cumC_keyworker[2], cumT_keyworker[2];
 	Infection *inf_queue[MAX_NUM_THREADS]; // the queue (i.e. list) of infections. 1st index is thread, 2nd is person.
 	int n_queue[MAX_NUM_THREADS]; 	// number of infections in inf_queue
+	HostClosure *host_closure_queue;  // When places close, host absenteeism times get set within threaded loop.
+	int host_closure_queue_size; // Number of hosts in a timestep to set min/max absenteeism times for.
```
igfoo, line 125: I think it would be good to explicitly reference the array that this gives the size of, i.e. `Number of host closures in host_closure_queue.`.

src/Sweep.cpp
```diff
@@ -1761,7 +1766,10 @@ int TreatSweep(double t)
 						} while (f3);
 					}
 
-			}
+			} // End of bs loop, tn loop, and prgama.
```
igfoo, line 1769: Typo `prgama`

src/Update.cpp
```diff
@@ -1246,6 +1251,23 @@ void DoPlaceClose(int i, int j, unsigned short int ts, int tn, int DoAnyway)
 	}
 }
 
+void UpdateHostClosure() {
+    // Single-threaded here, because the same host index might come up in
+	// different threads. (Which is the point of this buffered update...)
```
igfoo, line 1256: Whitespace is inconsistent here (looks like tabs vs spsaces)

igfoo, line 1256: This could be done multi-threaded, like `inf_queue`, but it's probably not worth the complexity unless profiling suggests it's a hotspot.

src/CovidSim.cpp
```diff
@@ -2969,6 +2972,11 @@ void SeedInfection(double t, int* nsi, int rf, int run) //adding run number to p
 			}
 		}
 	}
+
+	// Is there a possibility that place closure can happen in the DoInfect calls above?
+	// Not sure - but to play safe...
+	UpdateHostClosure();
```
igfoo, line 2978: I wonder whether, rather than having a number of calls to `UpdateHostClosure` after loops that use it, it would be possible / better to update it once per timestep in one of the loops in `RunModel` or something? I'm not familiar enough with the code to know, though...

src/Update.cpp
weshinsley: I don't think it'll be a hotspot - not enough work with just those two lines... 

I wondered about having all the threads traversing through the whole host_closure_queue together, but each of them only updating if (host_index % num_threads) == thread_no ...

weshinsley: Done!

src/Sweep.cpp
weshinsley: Well spotted...

src/Model.h
weshinsley: Yes, good point - comment changed.

src/Model.h
weshinsley: Done!

src/CovidSim.cpp
weshinsley: I wonder this too, and am also not quite familiar enough to be sure...

weshinsley: I'm taking this back to a draft, to just explore if (a) parallelism helps UpdateHostClosure, and (b) whether one call in RunModel makes any difference. After more staring at the code, I suspect it won't, but I'm going to rerun Report 9 results to be sure.

src/Update.cpp
weshinsley: Tested with parallelise and is was a bit slower! So reverted...

weshinsley:
Parallelisation actually slowed it down! Not enough work to overcome the fork overheads when using the modulo approach. Reverted to single-threaded UpdateHouseClosure, and retested against full Report 9 results.

src/CovidSim.cpp
weshinsley: Tested with just the one call in the RunModel loop, and replicated the report 9 results exactly.

weshinsley:
@igfoo - Thanks for comments - all addressed. @NeilFerguson if you have time for a quick look at this, just to check there is nothing I've missed; I think it's all fine, is replicating all the results perfectly, and still has no warnings in Intel...

weshinsley:
After discussion, going to do one more test with DCT switched on (which is not among the Report 9 tests), since there is a call to place closure in that part.

weshinsley:
Results using p_PC7_CI_HQ_SD_DCT.txt were identical when run with the latest commit on this branch, as with commit dfc97f5, which was just before I started the branch...

So I think we've done as much due diligence as possible, and we're good to go.

PR #376, opened 2020-06-08
Rename some parameters in Param.h

jieyouxu:
### Rename multiple parameters in `Param.h`.

Currently using `snake_case` for `struct Param` in `Param.h` because there is evidence of *multiple* naming conventions for different parameters: `PascalCase`, `camelCase`, `PascalCase_WithUnderscore`, etc., can be changed if there is a official code style guide.

### Refactor Scope

This PR is strictly restricted to *parameter naming changes*, and does not intend to refactor `struct Param` into smaller `struct`s and `enum`s. Future PRs are planned to split the work into multiple stages for easier review and testing. This PR should **not** change any type signatures and thus should **not** perform any semantic/behavioural changes.

This PR also attempts to minimize reformatting to minimize the diffs. A proper reformatting tool such as `clang-format` should be used as a follow-up PR to do an overall reformat due to changes in line length resulting from parameter name changes.

### TODOs

- Some parameter are not yet renamed, because:
    - I could not derive their intended **purpose** and **units of measurement** from existing documentation or from call sites / usage locations (some of them don't have documentation that I can find from `docs/` or from other issues / PRs).
    - Still debating on a good name for the parameter.

### Refactor Opportunities for Follow-up PRs

- `KernelType` really should be an `enum`. It seems that `KernelType` also determines which function should be used, so a compile-time function lookup table might be useful here.

### DIFF
--- a/src/Bitmap.cpp
+++ b/src/Bitmap.cpp
@@ -58,15 +58,15 @@ void CaptureBitmap()
 			}
 		}
 		logMaxPop = log(1.001 * (double)maxPop);
-		for (int i = 0; i < P.NMC; i++)
+		for (int i = 0; i < P.microcells_count; i++)
 			if (Mcells[i].n > 0)
 			{
 				f = 0;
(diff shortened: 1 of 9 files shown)


jieyouxu:
I only renamed a subset of the parameters, more renaming can happen in follow-up PRs.

Issue #188, opened 2020-05-08
Shadow variable


n3150n:
https://github.com/mrc-ide/covid-sim/blob/de9732e7715cfaa1a8ee6ba0aa85d13773b2b11a/src/Update.cpp#L812

Shadows variable in line 774.
Same happening in line 816

Feynstein:
@n3150n Can you edit your issue name with the current labels? I suggest [CS-xx][Very High] for this one

Feynstein:
Nevermind the numbering I forgot github had that

weshinsley:
Let's talk about this one on  #169, as there are more j-confusions before it. I think it is functionally fine, but rather unclear.

n3150n:
Agree.

n3150n:
already addressed in #169

Issue #169, opened 2020-05-07
Variable shadowing (j) in DoCase 


pulkin:
Disclaimer: I am in no way related to PVS Studio and their developers who kindly allowed checking arbitrary projects without buying their software.

Somebody dumped issue #165 into our work chat. I've run PVS studio (static analyser) on this repo just to make my own judgement on how bad (or good) the code quality is. Unfortunately, I did not build up any opinion on that regard but you guys might be interested in fixing these two `error`-level entries. Line numbers are offset by 2 or 3.

```bash
SetupModel.cpp	1058	err	V781 The value of the 'm' index is checked after it was used. Perhaps there is a mistake in program logic.
Update.cpp	799	err	V535 The variable 'j' is being used for this loop and for the outer loop. Check lines: 789, 799.
```

Besides these two, I think there is a massive potential for random memory corruption through `fscanf`, `sscanf` and `sprintf`, example (around 70 spotted throughout the code):

```bash
CovidSim.cpp  2421    warn    V512 A call of the 'sprintf' function will lead to overflow of the buffer 'outname'.
```

Please use `*nprintf`.

igfoo:
Thanks for the reports. It looks like
```
SetupModel.cpp	1058	err	V781 The value of the 'm' index is checked after it was used. Perhaps there is a mistake in program logic.
```
refers to
```
            denom_household[m]++;
            for (i2 = 0; i2 < m; i2++)
```

This doesn't look like a bug to me. First we record that there is a household of size `m`, and then we use `i2` to index each of the people in that household.

pulkin:
It is this line:

```c
while ((s > P.HouseholdSizeDistrib[ad][m - 1]) && (k + m < Mcells[j].n) && (m < MAX_HOUSEHOLD_SIZE)) m++;
```

I think the logic assumed by the analyser is that the condition `m < MAX_HOUSEHOLD_SIZE` checks the size of `P.HouseholdSizeDistrib[ad]`.

igfoo:
Ah, OK. Still looks like that one is a false positive, then.

[...]

PR #377, opened 2020-06-09
Refactor some variable shadowing

weshinsley:
This removes some variable shadowing observed in #169 and #188 - variable j is reused, the first one sharing the outer loop j, and two more that declare a new scoped j. 

There may be further ways to clean this up, and make it more clear in due course; for now I have only tried to make explicit what was previously happening (namely, exit value of j from the first loop being set). 

From discussions, there is no code issue here, just clarity.


dlaydon:
Looks good @weshinsley and thanks for fixing, just going to wait for checks to pass.

Issue #381, opened 2020-06-11
incDeath_X output contains cumDeath_X values?


bwynneHEP:
I've been looking at some of the severity data broken-down into the different age groups. I noticed a minor (but annoying) bug in one of the main outputs I plot: the incDeath_X output, where X takes values from 0 to 16 corresponding to the different age groups.

It seems that these incidence values output the exact duplicate of the cumulative values, and so they rise monotonically throughout a simulation. This does not appear to be true for incDeath_ILI_X, incDeath_SARI_X, or incDeath__Critical_X, and so one can presumably reconstruct the correct value by summing over these three outputs instead. Nonetheless, it would be helpful for the combined value output to be corrected.

I note that the outputs are created here:
https://github.com/mrc-ide/covid-sim/blob/master/src/CovidSim.cpp#L4083
using the `TSMean[i].incDa[j]` value.

Looking at where this value is set:
https://github.com/mrc-ide/covid-sim/blob/master/src/CovidSim.cpp#L5033
`TimeSeries[n].incDa[i] += (double)StateT[j].cumDa[i];` would appear to be the source of the mistake. However, I note many other values set in a similar way, so perhaps I have misunderstood the calculation?

This seems to only be an issue affecting the formatting of simulation output, and I doubt it would have affected anyone's results without them noticing.

dlaydon:
Hi @bwynneHEP and @weshinsley, yes you're right this does actually output the cumulative incidence. It was a quick hack that we post process in R. However you're right we should change it. Before I do though I need to check that it doesn't ruin anything upstream. Will get to it soon.

bwynneHEP:
Not an urgent problem - as you say, once I knew about it, fixing as a post-process step is easy.

Probably worth a careful look though, as there seemed to be a few incSomething values that were set a similar way (i.e. without the subtraction of the previous step value before the addition of the new one).

I also noticed that the corresponding cumDeath_X output is made by summing over the ILI, SARI and Critical categories at the moment the output is made inside `CovidSim.cpp`. This might mean that some other ad-hoc fixes have already been applied.

dlaydon:
[...]

PR #383, opened 2020-06-12
Dlaydon/add quarantined not sick

dlaydon:
Hi @matt-gretton-dann, I've added a couple of variables to the .severity.xls outputs, which will cause regression tests to fail. Otherwise output identical. Are you able to regenerate the target files then merge? Cheers!


weshinsley:
@dlaydon While you're doing this, I wonder whether it's worth looking at https://github.com/mrc-ide/covid-sim/issues/381

PR #384, opened 2020-06-13
math.h -> cmath

zdroid:
Follow up to #287, but without unnecessary changes.


weshinsley:
Are we sure this a literal "drop in" replacement on all platforms? (Including mac?) I'm a bit paranoid about the abs/fabs thing, so might do a full test suite to settle myself about it...

Issue #249, opened 2020-05-13
Alignment in Model.h


StevenWhitehouse:
I'm wondering what the reason was for the #pragma pack(push, 2) in Model.h. That seems an odd requirement for code that is presumably intended to be fast. Looking down the structures which are in that region, one in particular caught my eye, which is struct HOUSEHOLD. The reason is that if the compiler does align on a 2 byte basis the two floats in that struct are misaligned (i.e. 4 byte variables which will be on a 2 byte boundary).

Also, the field dum in struct POPVAR is never used, so can be removed. If that was added for alignment, then it is the wrong size, since CACHE_LINE_SIZE is in bytes, and with the two byte alignment being enforced, it is unlikely to produce the desired effect, although I didn't count the offset from the start of that struct to verify. It doesn't look like alignment should be needed at that point anyway.

I've copied & pasted a patch below with the proposed changes. While I was looking there are probably a couple of further things that can also be cleaned up (not included in the patch) which are:

- In struct POPVAR the nct_queue field is initialised, but otherwise appears unused, so possibly could be removed too. 
- Also the three bitfields towards the end of struct PERSON, namely digitalContactTraced, index_case_dct, and digitalContactTracingUser could be moved up next to the earlier set of bitfields (esocdist_comply, keyworker, to_die, detected) to save an unsigned int from that structure.

```
diff --git a/src/Model.h b/src/Model.h
index 801b185..218a209 100644
--- a/src/Model.h
+++ b/src/Model.h
@@ -11,8 +11,6 @@
 //// need to test that inequalities in IncubRecoverySweep can be replaced if you initialize to USHRT_MAX, rather than zero.
 //// need to output quantities by admin unit
 
-#pragma pack(push, 2)
-
 typedef struct PERSON {
        int pcell;                      // place cell, Cells[person->pcell] holds this person
        int mcell;                      // microcell, Mcells[person->mcell] holds this person
@@ -55,8 +53,8 @@ typedef struct PERSON {
 typedef struct HOUSEHOLD {
        int FirstPerson;
        unsigned short int nh; // number people in household
-       float loc_x, loc_y;
        unsigned short int nhr;
+       float loc_x, loc_y;
 } household;
 
 /*
@@ -111,7 +109,6 @@ typedef struct POPVAR {
        int* CellMemberArray, *CellSuscMemberArray;
        int** InvAgeDist;
        int* mvacc_queue;
-       int dum[CACHE_LINE_SIZE];
[...]

PR #386, opened 2020-06-15
Fix alignment in struct Household

StevenWhitehouse:
In Model.h there is a two byte packing pragma. In struct Household, that currently means
that the two floats are misaligned. That is based on int being 32 bits, and short being
16 bits, and float being 32 bits.

There is a simple solution to this, by pairing up the two short fields, that makes 32
bits and the two floats are then properly aligned. It may be possible to remove the
two byte packing pragma too, however this will affect other structures and I've not
yet reviewed those. So this a minimal change to fix this one issue.

Note that this will make snapshots incompatible from previous versions.

### DIFF
--- a/src/Model.h
+++ b/src/Model.h
@@ -58,8 +58,8 @@ struct Household
 {
 	int FirstPerson;
 	unsigned short int nh; // number people in household
-	float loc_x, loc_y;
 	unsigned short int nhr;
+	float loc_x, loc_y;
 };
 
 /*


StevenWhitehouse:
This is based on discussion in Issue #249 and it seems from comments there that the snapshot compatibility is not an issue. I did check and that seems to be the only time that struct Household is used in I/O. There are also potentially further changes that can be made, but in the interests of going one small step at a time, this seemed like a good starting point.

weshinsley:
As in Issue #249, this looks fine!

PR #390, opened 2020-06-16
Cleanup update bitmap output math

CloneDeath:
I primarily refactored some vector math in determining the pixel location to set.

I also added some explicitly typed operator overloading in order to mimic C++'s implicit casting of types, so we can multiple Vector2<double> with a Vector2<float> and the result will be a Vector2<double>. I'll be expanding on this as-needed, since the number of permutations is pretty large.

### DIFF
--- a/covid-sim.vcxproj
+++ b/covid-sim.vcxproj
@@ -104,6 +104,7 @@
     <ClCompile Include="src\Param.cpp" />
     <ClCompile Include="src\MicroCellPosition.cpp" />
     <ClCompile Include="src\Direction.cpp" />
+    <ClCompile Include="src\Geometry\Vector2.cpp" />
   </ItemGroup>
   <ItemGroup>
     <ClInclude Include="src\BinIO.h" />
--- a/src/Geometry/CMakeLists.txt
+++ b/src/Geometry/CMakeLists.txt
@@ -1,6 +1,7 @@
 target_sources(CovidSim
     PRIVATE
         ${CMAKE_CURRENT_LIST_DIR}/Size.h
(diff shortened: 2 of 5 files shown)


CloneDeath:
>     1. Newline at end of Vector2.cpp please (boring...!)
Fixed. Good eye! I'll be more careful about this in the future.

>     2. I wonder if a typedef for `Vector2<int>` as Pixel or Point might be helpful...?
I considered as much... I'd like to see where else we do Pixel Position math, and refactor that code together. I'll do some more research on this, and make a ticket. But for now, I think I'd like to keep the scope of this PR from getting too large.
 
>     3. [Edit - oops, I see you commented on this earlier] It's indeed tempting to go a bit further and factor out the common code in these 9 cases - they differ in the end only by which {one or two} out of {bmRecovered[j], bmInfected[j], bmTreated[j]} get incremented or deceremented. Eg something like:
Ha, yes. I 100% agree... but a future PR to keep this one focused and small... and to give me more time to think of a good solution. I'll go ahead and make a ticket for this though, so we don't forget about it.

weshinsley:
Arg, one more thing - kind of one that says more about me, but could we keep the extra bracket in;
`Vector2<int> pixel((Households[Hosts[ai].hh].loc * P.scale) - P.bmin);`
rather than
`Vector2<int> pixel(Households[Hosts[ai].hh].loc * P.scale - P.bmin);`

I know it's not necessary, but fatigue is a terrible thing, and it's worth the extra clarity sometimes.

CloneDeath:
Done. No problem at all :smiley:. I definitely see where you are coming from, and considered leaving them in.

PR #395, opened 2020-06-17
Improve if checks in Bitmap.cpp

zdroid:


zdroid:
@weshinsley macOS test was cancelled, why is that?

weshinsley:
I think a couple of PRs have to go through more urgently this morning. I'll restart tests when things are calmer...

PR #396, opened 2020-06-18
Added Cell/Hosts helper functions to Update.cpp

### LINKED ISSUE #339: Introduce State design pattern for Infection Status and Severity
Linked with #208 (Replace monolithic design with modular code)

InfStat and Severity are defined as enums, and transitions are implemented with functions in Update.cpp.

This could be refactored using the State design pattern, with a class for each infection status and severity, and clearly defined state transitions.

mstevens-uk:
Consolidate Cells and Hosts updates across Update.cpp into a helper function.

All changes are in Update.cpp.
A future step can extract this logic out of Update.cpp

assert statements (now removed) have been used to verify that the new UpdateCell() function gives the same result as the original logic.

Fixes #339

### DIFF
--- a/src/Update.cpp
+++ b/src/Update.cpp
@@ -25,9 +25,8 @@ void InfectiousToRecovered(int cellIndex);
 void InfectiousToDeath(int cellIndex);
 
 // if the dest cell state == src cell state, use this
-void UpdateCell(int* cellPeople, int destIndex, int srcIndex);
-
-void UpdateCell(int* cellPeople, int* srcCellPeople, int destIndex, int srcIndex);
+void UpdateCell(int* cellPeople, int index, int srcIndex);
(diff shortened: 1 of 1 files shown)


src/Update.cpp
```diff
@@ -23,6 +23,10 @@ void LatentToInfectious(int cellIndex);
 void InfectiousToRecovered(int cellIndex);
 void InfectiousToDeath(int cellIndex);
 
+// if the dest cell state == src cell state, use this
```
matt-gretton-dann, line 26: Can we have some more detailed comments on the parameters here please?  (Use doxygen formatting - I'd like to slowly add that as we go along - and VS 16.6 now supports that in tooltips).

mstevens-uk: Done.

PR #397, opened 2020-06-18
Add code formatting infrastructure

matt-gretton-dann:
This PR adds a .clang-format file to format code at the top-level and enforces it using a GitHub Action.

On Windows modern versions of Visual Studio should pick up the .clang-format file as well and automatically format code when saved.

The formatting choices made are based on Microsoft's defaults for Visual Studio.  Of interest though will be the choices between tabs & spaces, indent width and max line length.

I have chosen tabs with an indent of 2 spaces per-tab.  This is what we've had previously, but I'm interested in any other thoughts @NeilFerguson, @dlaydon, @weshinsley.

On line length - I've picked 100, as that is what the Linux Kernel has recently moved to, and my view is that I start losing context on longer lines.  Again up for discussion.


zdroid:
As I said in the previous thread: tabs make changelogs look bad by default and thus require additional user work. Unless you customize the tab size in Git diffs, they won't be displayed as you intended.

PR #399, opened 2020-06-18
Abstract cn_{1,2,3,4} functionality

zdroid:


src/Bitmap.cpp
```diff
@@ -303,10 +286,11 @@ void InitBMHead()
 	char buf[1024+3];
 	sprintf(buf, "%s.ge", OutFileBase);
 #ifdef _WIN32
```
matt-gretton-dann, line 288: Can we split this change out into a `create_directory()` function with platform-dependent implementations.  I don't like the `if` statement being platform specific.

zdroid: I think I didn't intend it originally for this PR. I reverted it here so I'll probably just make a new PR for that change.

src/Bitmap.cpp
```diff
@@ -123,27 +123,10 @@ void OutputBitmap(int tp)
...
-		case 3:
-			j = cn4;
-			cn4++;
-			sprintf(OutF, "%s.ge" DIRECTORY_SEPARATOR "Max.%s", OutFile, OutBaseName);
-			break;
+	for (auto &i : cn) {
+		j = i;
+		i++;
+		sprintf(OutF, "%s.ge" DIRECTORY_SEPARATOR "%s", OutFile, OutBaseName);
 	}
```
weshinsley, line 131: Where have the "Mean, Min, Max" gone from the filenames?

weshinsley: I'm also not totally clear on the "for auto" syntax - but don't you basically want something like...

j = cn[tp];
cn[tp]++;
sprintf(OutF, "%s.ge" DIRECTORY_SEPARATOR "%s.%s", OutFile, PreFix[tp], OutBaseName)

where PreFix is an array of {"", "Mean", "Min", "Max"}

zdroid: @weshinsley I don't know what was the intention for the original syntax, I just abstracted it. `for auto` seems the most appropriate for the given situation.

You're absolutely correct about the mean min max part, I missed to automate it. Do you want to implement the syntax (that you proposed) yourself, or want me to implement it instead?

weshinsley: In the original code, there's no loop; tp comes in as an argument to the function, and one of the cases in the switch gets called. In your version, you have a loop and all of them will get called won't they?

zdroid: Right, I'll fix the errors you noted.

zdroid:
@weshinsley Done.

zdroid:
I'm fairly confused by the build failing. Even more curiously, they didn't fail for the previous code version, which was incorrect as @weshinsley noted.

ozmorph:
> I'm fairly confused by the build failing. Even more curiously, they didn't fail for the previous code version, which was incorrect as @weshinsley noted.

I believe it's because your code increments `tp` instead of incrementing `cn[tp]`.

zdroid:
@ozmorph Thanks, fixed! Still doesn't explain how the previous tests actually passed. o.O

weshinsley:
Looks like we only do tests producing BMP output for tp=0 - so paths that needed the min./max./mean. parts of the filename weren't followed. And also:-

for (auto &i : cn)

Can't be used for simple arrays, only for C++ vectors etc, which have defined begin and end functions... Hence I suspect the loop was only executed once with i=cn[0]

zdroid:
@weshinsley 
```cpp
int cn[4] = {1, 2, 3, 4};
for (auto &i : cn) {
  i += 1;
}
for (auto i : cn) {
  std::cout << i;
}
```
prints out 2345. Proper use of reference operators in the for auto can be difficult, but it still works properly.

weshinsley:
I'll have a think about why the previous tests passed. It might actually have been that you were writing the same filename with the same contents multiple times, combined with the fact that we don't do regression tests on all the different kinds of bitmap that can be produced. (We do test the data behind such bitmaps though).

Anyway, latest version is simple and equivalent to the original, so have merged.

PR #388, opened 2020-06-15
US Regression Test Optimization

ozmorph:
Following up on [my comment](https://github.com/mrc-ide/covid-sim/pull/366#pullrequestreview-428065148) in #366 , I found low-hanging fruit inside the `AssignPeopleToPlaces()` function and MicroCellPosition code that results in a substantial (~42.8%) run-time reduction for the US regression test.

Merging this PR would allow for more frequent local testing.

Running `time ./regressiontest_US_based.py` on `master` on my workstation:

```
real    4m24.706s
user    4m52.256s
sys     0m7.381s
```

Running the same test on this PR:

```
real    2m31.731s
user    2m58.940s
sys     0m7.318s
```

PR #400, opened 2020-06-18
Reduce the number of for loops in CovidSim.cpp

zdroid:


src/CovidSim.cpp
```diff
@@ -705,14 +704,14 @@ void ReadParams(char* ParamFile, char* PreParamFile)
...
-		for (i = 0; i < NUM_AGE_GROUPS; i++)
 			P.PropAgeGroup[0][i] /= t;
-		t = 0;
-		for (i = 0; i < NUM_AGE_GROUPS; i++)
+
+			t = 0;
 			if (P.AgeSusceptibility[i] > t) t = P.AgeSusceptibility[i];  //peak susc has to be 1
-		for (i = 0; i < NUM_AGE_GROUPS; i++)
 			P.AgeSusceptibility[i] /= t;
+		}
```
matt-gretton-dann, line 714: I think this changes behaviour.

Loop at lines 706-708, tells us how many people there are in the age group.
Loop 709-710, then scales everything
Loop 712-713 finds the max age susceptibility
Loop 714-715 then scales everything.

src/CovidSim.cpp
```diff
@@ -733,19 +732,11 @@ void ReadParams(char* ParamFile, char* PreParamFile)
...
-			}
-			else
-			{
-				for (i = 0; i < NUM_AGE_GROUPS; i++)
-					for (j = 0; j < NUM_AGE_GROUPS; j++)
-						P.WAIFW_Matrix[i][j] = 1.0;
-			}
+					if (t > 0) P.WAIFW_Matrix[i][j] /= t;
+					else P.WAIFW_Matrix[i][j] = 1.0;
+				}
```
matt-gretton-dann, line 739: Again I think this changes behaviour.

src/CovidSim.cpp
zdroid: You are right. I missed that the last `t` from the first loop is used in the next loops. I'll correct those changes.

zdroid:
@matt-gretton-dann I merged two new for loops and removed the two problematic changes.

src/CovidSim.cpp
```diff
@@ -896,10 +894,8 @@ void ReadParams(char* ParamFile, char* PreParamFile)
...
 				P.InvJourneyDurationDistrib[i] = j;
-			}
-			for (i = j = 0; i <= 1024; i++)
-			{
-				s = ((double)i) / 1024;
+
+				j = 0;
 				while (P.LocalJourneyDurationDistrib[j] < s)j++;
 				P.InvLocalJourneyDurationDistrib[i] = j;
 			}
```
weshinsley, line 901: This will be much slower, as you're resetting `j` in each iteration. `j` is stepping through, I think, cumulative distributions (see their calculation a bit earlier) so you can leave `j` where it is and continue searching onwards.

If you *really* want to merge the loops, I think you need...
for (i = j1 = j2 = 0; i <= 1024; i++) 
{
    s = ((double)i) / 1024;
    while (P.JourneyDurationDistrib[j1] < s) j1++;
    P.InvJourneyDurationDistrib[i] = j1;
    while (P.LocalJourneyDurationDistrib[j2] < s) j2++;
    P.InvLocalJourneyDurationDistrib[i] = j2;
}

zdroid: Can you benchmark that somehow? On the other side, the code is looped 1024 instead of 2048 times, and on top of that, i isn't calculated twice. I think the benefits of that are pretty clear.

ozmorph: As @weshinsley pointed out, `P.JourneyDurationDistrib` and `P.LocalJoureyDurationDistrib` are both cumulative distribution functions (i.e. the numbers will be ordered from smallest to largest). If you reset `j = 0` you are repeating unnecessary comparisons up until the previous `j` value search since `P.JourneyDurationDistrib` at every value before `j` will be less than `(i + 1) / 1024`. The outer loop may have double the iterations, but the inner loop is what will be calculated the most if `j` were reset every time.

I think that the merged loop is the correct refactor here. It halves the number of outer loop iterations while keeping the number of inner loop iterations the same. The only cost is an additional integer variable. However to be completely frank, this is a premature optimization. I've been profiling the code this week (see #388) and my analysis shows that the entire `ReadParams()` function takes less than 0.1% of the run-time.

weshinsley: Yes, I agree, this section isn't one to overthink about performance - and also I don't think this PR was intended to be about performance - it's really a bite-sized code clean-up, where the review highlighted ways it behaved differently (on both behaviour and performance) to the original.

But all looks good to me now.

matt-gretton-dann: > Is that solution slower than the current one (before my change)? I would proceed with using that.

As a serious answer to this question: Who knows?

These loop merging PRs concern me because they change memory access patterns - and therefore how the CPU will feed the "beast" that is the cache and memory hierarchy.

The original loops were simple - two arrays being accessed sequentially in each loop.  I would expect a server-class CPU to be able to handle this and be able to track the accesses patterns to request memory far enough ahead of time that we won't get stalls in the memory system.  However, merging these loops means we're interleaving more accesses.  I would expect some CPUs to start to struggle with this and the code to actually be slower.

CovidSim's biggest performance bottle-neck is memory accesses - real runs use far more memory than available in the caches.  So giving the CPU all the help it can to predict memory usage is likely to be best way to improve performance.  This involves things like thinking about structure layout, data locality, and access patterns.

Having said that - in this case we're doing initialisation in cold code so performance is not as much an issue as readibability.

zdroid: @matt-gretton-dann If this change is problematic, I can propose the one at lines 675 only.

matt-gretton-dann: No I'm fine with this as it is - I'm just trying to raise awareness that even simple changes can/will have significant knock on effects around the performance of this code.

PR #402, opened 2020-06-18
Simplify Param::get_micro_cell_index_from_position

zdroid:
If the function doesn't have some hidden logic behind it, then it can be simplified based on the fact that
(a + b) % b = a % b + b % b = a % b


matt-gretton-dann:
The magic is that there is nowhere else ensures that the indices in `position` are valid.  The original function ensures that values in the range [-b, +inf] get mapped to [0, b).

zdroid:
Maybe not the best solution to do that, but I understood. Closing.

PR #404, opened 2020-06-19
Added Severity helper functions to Update.cpp

mstevens-uk:
Consolidate Severity StateT updates into a set of  helper functions.
There are 2 functions for each severity state (used) - for transitioning from and to.
There is then a common function ChangeSeverity which increments/decrements the state variables.

All changes are in Update.cpp.
The next step is to extract these helper functions out of Update.cpp.

### DIFF
--- a/src/Update.cpp
+++ b/src/Update.cpp
@@ -18,30 +18,35 @@ void RecordEvent(double, int, int, int, int); //added int as argument to InfectS
 unsigned short int ChooseFromICDF(double *, double, int);
 Severity ChooseFinalDiseaseSeverity(int, int);
 
-// state transition helpers
+// infection state transition helpers
 void SusceptibleToRecovered(int cellIndex);
 void SusceptibleToLatent(int cellIndex);
 void LatentToInfectious(int cellIndex);
 void InfectiousToRecovered(int cellIndex);
 void InfectiousToDeath(int cellIndex);
 
+// severity state transition helpers
 
-void ToInfected(int tn, Person* a, int ai, double q);
-void FromMild(int tn, Person* a, int ai);
-void ToMild(int tn, Person* a, int ai);
-void FromCritRecov(int tn, Person* a, int ai);
+void ToInfected(int tn, short infectType, int ai, double q);
 
-void FromSARI(int tn, Person* a, int ai);
+void FromMild(int tn, int microCellIndex, int ai);
+void ToMild(int tn, int microCellIndex, int ai);
 
-void FromILI(int tn, Person* a, int ai);
-void ToILI(int tn, Person* a, int ai);
+void FromCritRecov(int tn, int microCellIndex, int ai);
 
-void ILIToSARI(int tn, Person* a, int ai);
-void SARIToCritical(int tn, Person* a, int ai);
-void CriticalToRecovered(int tn, Person* a, int ai);
-void CriticalToDeath(int tn, Person* a, int ai);
-void SARIToDeath(int tn, Person* a, int ai);
-void ILIToDeath(int tn, Person* a, int ai);
+void FromSARI(int tn, int microCellIndex, int ai);
+
+void FromILI(int tn, int microCellIndex, int ai);
+void ToILI(int tn, int microCellIndex, int ai);
+
+void ILIToSARI(int tn, int microCellIndex, int ai);
+void ILIToDeath(int tn, int microCellIndex, int ai);
+
+void SARIToCritical(int tn, int microCellIndex, int ai);
+void SARIToDeath(int tn, int microCellIndex, int ai);
+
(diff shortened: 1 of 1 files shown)


src/Update.cpp
```diff
@@ -355,48 +326,26 @@ void DoDeath_FromCriticalorSARIorILI(int ai, int tn)
...
+				ToDeathSARI(tn, a->mcell, ai);
+				break;
+
+			case Severity_ILI:
+				FromILI(tn, a->mcell, ai);
+				ToDeathILI(tn, a->mcell, ai);
+				break;
 		}
+
+		//// change current status (so that flags work if function called again for same person). Don't move this outside of this if statement, even though it looks like it can be moved safely. It can't.
```
dlaydon, line 347: Thank you @mstevens-uk for this. Overall, it tidies up the movement between the severity states well, and is appreciated greatly. However, there is a warning explicitly not to do this: please put the assignment `a->Severity_Current = Severity_Dead;` back into the individual cases of this switch. The switch does not consider all severity cases (i.e. is not exhaustive), so in rare cases `DoDeath_FromCriticalorSARIorILI` can be called before a person has had their severity assigned. Therefore they will die before they onset with symptoms - this causes all kinds of problems with bookkeeping (total deaths don't add up). It will pass regression tests because the ILI CFR is set to impossible in the example pre-parameters, but this won't always be so, and because ILI cases are so common this can matter a great deal. An identical problem can happen with `DoRecover_FromSeverity`  but I can see that this part of your code changes is fine.

mstevens-uk: @dlaydon Thanks for your review. Yes, good spot. I've put the assignment back into the switch cases, and added the extra detail from your review into the comments.

src/Update.cpp
```diff
@@ -18,13 +18,30 @@ void RecordEvent(double, int, int, int, int); //added int as argument to InfectS
...
+void FromSARI(int tn, int microCellIndex, int ai);
+void ToSARI(int tn, int microCellIndex, int ai);
+void FromILI(int tn, int microCellIndex, int ai);
+void ToILI(int tn, int microCellIndex, int ai);
+void FromCritical(int tn, int microCellIndex, int ai);
+void ToCritical(int tn, int microCellIndex, int ai);
+void ToDeathILI(int tn, int microCellIndex, int ai);
+void ToDeathSARI(int tn, int microCellIndex, int ai);
+void ToDeathCritical(int tn, int microCellIndex, int ai);
+
```
matt-gretton-dann, line 44: If only used in this file please use `static` or an anonymous namespace.

src/Update.cpp
```diff
@@ -1435,6 +1377,122 @@ void InfectiousToDeath(int cellIndex)
 	Cells[cellIndex].D++; //// one more dead person
 }
 
+// severity state functions
+
+void decrement(int& x) {x--;}
+void increment(int& x) {x++;}
```
matt-gretton-dann, line 1383: I would prefer that we use `std::function` and lambdas instead of function pointers.

Also I think this is an opportunity for some asserts/checks (that `x` never goes negative, either on the way down or on the way up due to overflow).

src/Update.cpp
```diff
@@ -1435,6 +1377,122 @@ void InfectiousToDeath(int cellIndex)
...
+void ToSeverity(int& quantity, int* age, int* adUnit, int microCellIndex, int ai)
+{
+	void (*fcnPtr)(int& x) { increment };
+
+	ChangeSeverity(quantity, age, adUnit, microCellIndex, ai, fcnPtr);
+}
+
+void ToInfected(int tn, short infectType, int ai, double q)
+{
+	///// Change threaded state variables to reflect new infection status of person a.
```
matt-gretton-dann, line 1414: Person `a` or person `ai`?

src/Update.cpp
mstevens-uk: Happy to add asserts for each severity state variable post-transition to check they are >= 0.
There are no asserts at present in CovidSim - do you want more of them going forward?

matt-gretton-dann: I would :-). Let's use assert to begin with - we can think about moving to GSL's precond/postcond/invariant macros as we go forward.

src/Update.cpp
mstevens-uk: helper functions are now static

src/Update.cpp
mstevens-uk: ai

src/Update.cpp
mstevens-uk: Done.

mstevens-uk:
@matt-gretton-dann @weshinsley @dlaydon PR #404 is ready to go. There is a clear mapping between the old increments/decrements of the severity state variables, and the new extracted functions. There should be no other logic changes. Matt had flagged that one function did break, but that has now been resolved.

Issue #403, opened 2020-06-19
Extracting classes from the C code


zebmason:
Whilst I've been contributing to this project I've be using it as a test case for enhancing my [dependency analysis tool](https://www.deepends.co.uk/) and have now shipped code to help extract classes from C code as shown on [YouTube](https://youtu.be/Fvl5XrcpSwo). Anyone wanting to play can download and use for free until the end of July. Hopefully by which time I'll have worked out how to make money from it as well as sorted a proper installer.

PR #405, opened 2020-06-19
Extract a struct for building bitmaps

zebmason:
First pass tidy up as seen in the video referenced in #403


src/Bitmap.cpp
```diff
@@ -24,36 +24,50 @@ static CLSID  encoderClsid;
 #include <sys/stat.h> // for mkdir
 #endif
 
-static unsigned char* bmf, *bmPixels, *bmp;
+namespace CovidSim
+{
+	namespace BitMap
```
matt-gretton-dann, line 29: Bikeshedding comment :-): Can it be `Bitmap` as it is not a `std::map<>` derived type which is what capitalising the M makes me think of!

zebmason: Tried Bitmap initially but it clashes with the class of the same name. I suppose it could be changed to Graphics or some such.

zebmason:
@matt-gretton-dann Just a first pass as I'm wary about trying to get changes accepted given that I've had several PRs just rotting for over a month leaving really bad merge conflicts to get resolved.

PR #195, opened 2020-05-08
Replace all malloc() with calloc()

chrisatang:
Tackles issue #77.

PR #406, opened 2020-06-19
Rework calls to memory routines

matt-gretton-dann:
This set of commits reworks memory routine usage.  It does two things:

 * Add x*alloc() routines which wrap the *alloc() routines and ensure that on errors we abort. This simplifies the calls to ERR_CRITICAL in a more debuggable way (we can stack trace through the abort call to find the specific failure).
 * Replace xmalloc() calls with xcalloc() calls (this is an updated form of #195.

### DIFF
--- a/src/Memory.cpp
+++ b/src/Memory.cpp
@@ -1,0 +1,74 @@
+/** \file  Memory.h
+ *  \brief Provide memory routines that terminate on failure
+ */
+
+#include "Memory.h"
+
+#include <cstdlib>
+#include <cstring>
+#include <cerrno>
+
+#include "Error.h"
+
+void* Memory::xmalloc(std::size_t size) noexcept
+{
+  /* Ensure we're going to allocate some memory.  */
+  if (size == 0)
+  {
+    size = 1;
+  }
+
+  void* result = std::malloc(size);
+  if (result == nullptr)
+  {
+    ERR_CRITICAL_FMT("Unable to allocate memory: %s\n", std::strerror(errno));
+  }
+
+  return result;
+}
+
+void* Memory::xcalloc(std::size_t nelem, std::size_t elsize) noexcept
+{
+  /* Ensure we're going to allocate some memory.  */
+  if (elsize == 0 || nelem == 0)
+  {
+    /* Memory we allocate shouldn't be used - so just set it to the smallest
+     * amount possible.  */
+    elsize = 1;
+    nelem = 1;
+  }
+
+  void* result = std::calloc(nelem, elsize);
+  if (result == nullptr)
+  {
+    ERR_CRITICAL_FMT("Unable to allocate memory: %s\n", std::strerror(errno));
+  }
+
+  return result;
+}
+
+void* Memory::xrealloc(void* ptr, std::size_t size) noexcept
+{
+  /* Ensure we're going to allocate some memory.  */
+  if (size == 0)
+  {
+    size = 1;
+  }
+
+  void* result = std::realloc(ptr, size);
+
+  if (result == nullptr)
+  {
+    /* On failure realloc doesn't free the previously allocated memory.  */
+    std::free(ptr);
+    ERR_CRITICAL_FMT("Unable to re-allocate memory: %s\n", std::strerror(errno));
+  }
+
+  return result;
+}
+
+void Memory::xfree(void* ptr) noexcept
+{
+  std::free(ptr);
+}
+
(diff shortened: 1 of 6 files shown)


src/Memory.cpp
```diff
@@ -0,0 +1,74 @@
...
+#include <cerrno>
+
+#include "Error.h"
+
+void* Memory::xmalloc(std::size_t size) noexcept
+{
+  /* Ensure we're going to allocate some memory.  */
+  if (size == 0)
+  {
+    size = 1;
```
ozmorph, line 18: I wonder if a warning should be printed to the log if not an error and exit. I can't off the top of my head think of a valid reason to request an allocation of 0 bytes. If that's the case, it would help to discover logic errors in the code rather than create a 1-byte heap allocation. My two cents at least.

matt-gretton-dann: This is not unusual as we've had instances where we're allocating `0` elements of something (validly) and the difference in platform behaviours (as this implementation defined behaviour) caused issues.

I do agree a logged comment would be useful, but that we then allow it to pass.

matt-gretton-dann:
> For later - do you think there is value some in sort of optional (#define TEST_MEMORY) automatic deconstruction at the end with these?

I did try this - but the sequence is complicated enough and with conditionals that I got confused and stopped with other higher priority things.  Which isn't to say this is a bad idea - getting a clean bill from `valgrind` is useful as then any reports it does give are definitely issues.

I think the ultimate solution is to move to C++ classes,use RAII, and will free their memory automatically as the objects are destructed.  This will take time though.

ZedThree:
What's the policy on templates in the code? Lots of the calls look like

```cpp
variable = (type*)Memory::xcalloc(nelems, sizeof(type));
```

which involves repeating the type, as well as a C-style cast. Adding something like:

```cpp
template <typename T>
T* Memory::xcalloc(std::size_t nelems) noexcept {
  return static_cast<T*>(Memory::xcalloc(nelems, sizeof(T)));
}
```

would make the calls look like:

```cpp
variable = Memory::xcalloc<type>(nelems);
```

and in some places can even avoid repeating the type three times by using `auto`:

```diff
-	char** AdunitNames, * AdunitNamesBuf;
-	if (!(AdunitNames = (char**)malloc(3 * ADUNIT_LOOKUP_SIZE * sizeof(char*)))) ERR_CRITICAL("Unable to allocate temp storage\n");
-	if (!(AdunitNamesBuf = (char*)malloc(3 * ADUNIT_LOOKUP_SIZE * 360 * sizeof(char)))) ERR_CRITICAL("Unable to allocate temp storage\n");
+	auto AdunitNames = Memory::xcalloc<char*>(3 * ADUNIT_LOOKUP_SIZE);
+	auto AdunitNamesBuf = Memory::xcalloc<char*>(3 * ADUNIT_LOOKUP_SIZE * 360);
```

matt-gretton-dann:
> What's the policy on templates in the code? Lots of the calls look like
> 
> ```c++
> variable = (type*)Memory::xcalloc(nelems, sizeof(type));
> ```
> 
> which involves repeating the type, as well as a C-style cast. Adding something like:
> 
> ```c++
> template <typename T>
> T* Memory::xcalloc(std::size_t nelems) noexcept {
>   return static_cast<T*>(Memory::xcalloc(nelems, sizeof(T)));
> }
> ```
> 
> would make the calls look like:
> 
> ```c++
> variable = Memory::xcalloc<type>(nelems);
> ```
> 
> and in some places can even avoid repeating the type three times by using `auto`:
> 
> ```diff
> -	char** AdunitNames, * AdunitNamesBuf;
> -	if (!(AdunitNames = (char**)malloc(3 * ADUNIT_LOOKUP_SIZE * sizeof(char*)))) ERR_CRITICAL("Unable to allocate temp storage\n");
> -	if (!(AdunitNamesBuf = (char*)malloc(3 * ADUNIT_LOOKUP_SIZE * 360 * sizeof(char)))) ERR_CRITICAL("Unable to allocate temp storage\n");
> +	auto AdunitNames = Memory::xcalloc<char*>(3 * ADUNIT_LOOKUP_SIZE);
> +	auto AdunitNamesBuf = Memory::xcalloc<char*>(3 * ADUNIT_LOOKUP_SIZE * 360);
> ```

Templates are fine.  However, I take `x*` routines as being simple wrappers around standard library functions with error handling and implementation defined behaviours 'smoothed out'.  This PR is that change.

I don't dislike your suggestions - however I wonder whether being a bit more piecemeal but radical in changes using standard library features (unique_ptr, shared_ptr and others) will be better.

zdroid:
@matt-gretton-dann Great change! You may be interested to close #195, since this supersedes it.

matt-gretton-dann:
Closed - superseded by #406

PR #324, opened 2020-05-22
Add in GoogleTest support

zebmason:

PR #407, opened 2020-06-20
Remove USE_WIN32_BMP from build file

zebmason:
Follow up to #324


matt-gretton-dann:
Thank you for doing this.

You missed the src/CMakeLists.txt use though (I found that by doing a `grep -r WIN32_BM .`).

zebmason:
@matt-gretton-dann Thought it would be quicker if I did the changes then you do the approval and merge. Given that I know I'm useless in the morning until after the second mug of tea it explains why I was getting confused trying to use Notepad++ first thing.

zebmason:
@matt-gretton-dann The define in src/CMakeLists.txt doesn't seem to do anything (anymore?) and the tests pass.

PR #408, opened 2020-06-20
Add in bounding box for double

zebmason:
Separate out the geometry stuff into a separately built library.
Add unit tests for the new library.
Remove some YAGNI code from Vector2 that was causing build issues.
Actually use bounding box code so it is no longer YAGNI.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -10,23 +10,22 @@ set(MAIN_HDR_FILES CovidSim.h BinIO.h Rand.h Constants.h Country.h Error.h
   InverseCdf.h Memory.h CLI.h)
 source_group(covidsim\\main FILES ${MAIN_SRC_FILES} ${MAIN_HDR_FILES})
 
-add_subdirectory(Geometry)
-
 # CovidSim target
 add_executable(CovidSim ${MAIN_SRC_FILES} ${MAIN_HDR_FILES})
 target_include_directories(CovidSim PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
 
+add_subdirectory(Geometry)
 add_subdirectory(Models)
 
+target_link_libraries(CovidSim PUBLIC geometrylib)
 if(USE_OPENMP)
-  target_link_libraries(CovidSim PUBLIC geometrylib OpenMP::OpenMP_CXX)
+  target_link_libraries(CovidSim PUBLIC OpenMP::OpenMP_CXX)
 endif()
 if(WIN32)
-  target_link_libraries(CovidSim PUBLIC geometrylib Gdiplus.lib Vfw32.lib)
+  target_link_libraries(CovidSim PUBLIC Gdiplus.lib Vfw32.lib)
   target_compile_definitions(CovidSim PUBLIC  "_CRT_SECURE_NO_WARNINGS")
   add_compile_options("fp:strict")
 elseif(UNIX)
-  target_link_libraries(CovidSim PUBLIC geometrylib)
   target_compile_definitions(CovidSim PUBLIC UNIX)
 else()
   message(FATAL_ERROR "Unknown operating system type: ${CMAKE_SYSTEM_NAME}")
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -10,21 +10,23 @@ set(MAIN_HDR_FILES CovidSim.h BinIO.h Rand.h Constants.h Country.h Error.h
   InverseCdf.h Memory.h CLI.h)
 source_group(covidsim\\main FILES ${MAIN_SRC_FILES} ${MAIN_HDR_FILES})
 
+add_subdirectory(Geometry)
+
 # CovidSim target
 add_executable(CovidSim ${MAIN_SRC_FILES} ${MAIN_HDR_FILES})
 target_include_directories(CovidSim PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
 
-add_subdirectory(Geometry)
 add_subdirectory(Models)
 
 if(USE_OPENMP)
-  target_link_libraries(CovidSim PUBLIC OpenMP::OpenMP_CXX)
+  target_link_libraries(CovidSim PUBLIC geometrylib OpenMP::OpenMP_CXX)
 endif()
 if(WIN32)
-  target_link_libraries(CovidSim PUBLIC Gdiplus.lib Vfw32.lib)
+  target_link_libraries(CovidSim PUBLIC geometrylib Gdiplus.lib Vfw32.lib)
   target_compile_definitions(CovidSim PUBLIC  "_CRT_SECURE_NO_WARNINGS")
   add_compile_options("fp:strict")
 elseif(UNIX)
+  target_link_libraries(CovidSim PUBLIC geometrylib)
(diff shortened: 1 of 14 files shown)


src/CMakeLists.txt
```diff
@@ -5,25 +5,27 @@ set(MAIN_SRC_FILES CovidSim.cpp BinIO.cpp Rand.cpp Error.cpp Dist.cpp Kernels.cp
...
 # CovidSim target
 add_executable(CovidSim ${MAIN_SRC_FILES} ${MAIN_HDR_FILES})
 target_include_directories(CovidSim PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
 
-add_subdirectory(Geometry)
 add_subdirectory(Models)
 
 if(USE_OPENMP)
-  target_link_libraries(CovidSim PUBLIC OpenMP::OpenMP_CXX)
+  target_link_libraries(CovidSim PUBLIC geometrylib OpenMP::OpenMP_CXX)
```
matt-gretton-dann, line 17: I think multiple `target_link_libraries` calls are additive - so we only need one here.

src/CMakeLists.txt
```diff
@@ -5,25 +5,27 @@ set(MAIN_SRC_FILES CovidSim.cpp BinIO.cpp Rand.cpp Error.cpp Dist.cpp Kernels.cp
 set(MAIN_HDR_FILES CovidSim.h BinIO.h Rand.h Constants.h Country.h Error.h Dist.h Kernels.h Bitmap.h Model.h Param.h SetupModel.h ModelMacros.h InfStat.h CalcInfSusc.h Sweep.h Update.h MicroCellPosition.hpp Direction.hpp InverseCdf.h)
 source_group(covidsim\\main FILES ${MAIN_SRC_FILES} ${MAIN_HDR_FILES})
 
+add_subdirectory(Geometry)
```
matt-gretton-dann, line 8: Can we put all `add_subdirectory()` calls together as far as possible.  I don't think the ordering here is important.

src/Geometry/BoundingBox.h
```diff
@@ -1,43 +1,187 @@
...
-		       && point.y < end.y;
-	}
-
-	template<class T>
-	Size<T> BoundingBox<T>::size() const {
-		return Size<T>(this->width(), this->height());
-	}
+  /// General class for representing a 2D maximum point
+  template<typename T>
+  class Maximum2 : public Vector2<T>
```
matt-gretton-dann, line 13: Maximum of what?  My instant reading was (as we're talking about vectors) this was a vector with maximum length (x*x + y*y).  Actually its the top-right corner of a bounding box (and Minimum is the bottom-left).  So I wonder if there are better names.

src/Geometry/BoundingBox.h
```diff
@@ -1,43 +1,187 @@
...
+  {
+  public:
+    /// Default constructor initialises nothing
+    Maximum2() = default;
```
matt-gretton-dann, line 17: I would be tempted to not have a default constructor as that leaves us in an unknown state.  We should move to having classes always in a valid state (even if that is an explicit uninitialised state).

src/Geometry/BoundingBox.h
```diff
@@ -1,43 +1,187 @@
...
+    /// \param y Y coordinate
+    Maximum2(T x, T y)
+      : Vector2<T>(x, y)
+    {
+    }
+
+    /// Adjust so it is the maximum of this and the other point
+    /// \param p Point
+    /// \param offset Value to add to component of p
+    void add(const Vector2<T>& p, T offset)
```
matt-gretton-dann, line 30: I think add is the wrong name here.  Something more like "expand" or "include" as we're expanding the geometry to include a point. in its bounding box.

unit_tests/geometry/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
...
+
+FIND_PACKAGE(Threads REQUIRED)
+
+# from list of files we'll create tests test_name.cpp -> test_name
+foreach(_test_file ${TEST_GEOMETRY_SRC_FILES})
+    get_filename_component(_test_name ${_test_file} NAME_WE)
+    add_executable(${_test_name} ${_test_file})
+    target_link_libraries(${_test_name} geometrylib gtest_main ${CMAKE_THREAD_LIBS_INIT})
+    gtest_discover_tests(${_test_name})
+endforeach()
```
matt-gretton-dann, line 15: Lots of this feels like repeated boiler-plate that will be used by other tests can we add a function() to call to add tests?

src/Geometry/BoundingBox.h
matt-gretton-dann: Ah I wonder if the ultimate issue is that `Vector` isn't being used as a `Vector` here its being used as a `Position`.  A vector is a direction and distance, it doesn't specify a start or end point.  Whereas, what we're doing is specifying a bounding box with a particular place in my for (0, 0).

Vectors and Positions mostly have a similar interface so the fix may be: `template<typename T> using Position<T> = Vector<T>;` or some such.

zebmason: It is a point so a vector offset from the origin (0, 0). Alternatively I am used to dealing with homogeneous points (x, y, z, w) for which a position is of the form (x, y, z, 1) and a displacement is (x, y, z, 0) so that when it is transformed by a 4 by 4 matrix a position will be rotated and translated whereas a displacement will only be rotated.

I'll try and make it a bit clearer in the code.

src/CMakeLists.txt
zebmason: You would appear to be correct as the tests pass. Thanks.

src/CMakeLists.txt
zebmason: Moved back into its original position as the geometry binary isn't added to the exe until further down the file.

src/Geometry/BoundingBox.h
zebmason: Just uses the default constructor from Vector2<T> now.

src/Geometry/BoundingBox.h
zebmason: I prevaricated over a while as add seems to imply we are building a set of points rather than altering the bounds. Changed to expand even though the point added will not necessarily expand the bounds. Couldn't think of an appropriate word but could have written a sentence in snake case which would have been rather pointless.

unit_tests/geometry/CMakeLists.txt
zebmason: Good point, there is now a function.

src/Geometry/BoundingBox.h
zebmason: I've reviewed this now and think that the comments make it clear that the vectors are not displacement vectors but points.

zebmason:
Note that this is one of the final bits to be extracted from #213 that dates back to 11th May the other bits being the upgrading of CMake and the adding of GoogleTest

zebmason:
@matt-gretton-dann Any reason to keep rebasing this?

weshinsley:
We had to pause everything to carefully merge some new science work, which took a little time...

zebmason:
@weshinsley The rebase wasn't too bad but I did notice the reason for the conflicts. This PR gives you code with unit tests and documentation that are effectively two months old.

PR #411, opened 2020-06-20
Fix issue with initialisation of inv journey dists

matt-gretton-dann:
I noticed this whilst reviewing #400, and I'm not sure if it is a bug or not - @NeilFerguson I would appreciate your review.

I believe it **doesn't** affect any simulations published to date as it requires airports to be enabled - which they haven't been for CovidSim.

In the two loops starting at line 894 of src/CovidSim.cpp initialising `InvJourneyDurationDistrib` and `InvLocalJourneyDurationDistrib`, the variable `j` is initialised to `0` at the head of the loop and then used to find an appropriate index in the `while` loop, which is used to set the value of `Inv...[i]`.

The issue is `j` is not reset each go round the loop.  Therefore `Inv...[i - 1] <= Inv...[i]` (j is always increasing).

I am not sure whether this is a bug or not.  I think it probably is and so have made the proposed change - but @NeilFerguson can you please confirm?

### DIFF
--- a/src/CovidSim.cpp
+++ b/src/CovidSim.cpp
@@ -891,14 +891,16 @@ void ReadParams(char* ParamFile, char* PreParamFile)
 				P.JourneyDurationDistrib[i] += P.JourneyDurationDistrib[i - 1];
 				P.LocalJourneyDurationDistrib[i] += P.LocalJourneyDurationDistrib[i - 1];
 			}
-			for (i = j = 0; i <= 1024; i++)
+			for (i = 0; i <= 1024; i++)
 			{
+				int j = 0;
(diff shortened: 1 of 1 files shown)


weshinsley:
I wonder if P.JourneyDurationDistrib and P.LocalJourneyDurationDistrib are changing monotonically - hence you can always resume j where you previously left off, in order to find the right value...

matt-gretton-dann:
@weshinsley : Thanks for the review.  So I'll "correct" the code by adding a comment that this is what is going on, and leave the code as originally written.

weshinsley:
This was refactored at some point to make it more clear...

PR #412, opened 2020-06-21
Replace counter assignment with breaks

zdroid:


zdroid:
In the second example goto would be more idiomatic. In the first it's clear since there is only one loop. I found the current syntax very confusing, as I had to re-check the initial loop conditions to see that it'd end.

The way that I proposed it the loop ends directly, instead of changing a variable and then checking its value to end the loop.

zdroid:
@matt-gretton-dann I'm missing something. In what way is that PR related to this? The lines that I changed here are untouched in it.

ozmorph:
> @matt-gretton-dann I'm missing something. In what way is that PR related to this? The lines that I changed here are untouched in it.

They may have accidentally have been referring to my other PR #346 which (once complete) will remove the specified function entirely.

PR #414, opened 2020-06-22
Replace P.clPx with P.clP[x]

zdroid:


matt-gretton-dann:
This now needs reworking with the changes from #228 merged into master.

zdroid:
@matt-gretton-dann Done.

src/CovidSim.cpp
```diff
@@ -5517,66 +5517,16 @@ int GetInputParameter3(FILE* dat, const char* SItemName, const char* ItemType, v
...
-					else if (n == 2)
-						* ((int*)ItemPtr) = (int)P.clP1;
-					else if (n == 3)
-						sscanf(match, "%s", (char*)ItemPtr);
-				}
-				else if ((match[0] == '#') && (match[1] == '2'))
-				{
+				if (match[0] == '#') {
 					FindFlag++;
+					int match_id = match[1] - '1';
```
weshinsley, line 5522: If you do it this way, should check that match[1] is in range. 
The previous version would have safely ignored anything outside of 1-6, so let's keep it the same.

```
if (match[0] == '#') {
  int match_id = match[1] - '1';
  if ((match_id >= 1) && (match_id <= 6)) {
    FindFlag++;
       ... the if n stuff ...
  }
}
```

zdroid: @weshinsley Fixed.

PR #415, opened 2020-06-22
Use enum struct for Direction

zdroid:
This disallows potentially unsafe and unidiomatic behavior such as comparision of enums to integers. It also avoids the usage of common names such as Up and Down in the global namespace.


zdroid:
rebased

PR #416, opened 2020-06-22
Use enum struct for InfStat

zdroid:
I think `InfStat` can be reworked to avoid using `abs(InfStat::...)`, but I'd rather leave that for some next PR.

### DIFF
--- a/src/CovidSim.cpp
+++ b/src/CovidSim.cpp
@@ -2489,7 +2489,7 @@ void InitModel(int run) // passing run number so we can save run number in the i
 			Hosts[k].detected = 0; //set detected to zero initially: ggilani - 19/02/15
 			Hosts[k].detected_time = 0;
 			Hosts[k].digitalContactTraced = 0;
-			Hosts[k].inf = InfStat_Susceptible;
+			Hosts[k].inf = InfStat::Susceptible;
 			Hosts[k].num_treats = 0;
 			Hosts[k].latent_time = Hosts[k].recovery_or_death_time = 0; //also set hospitalisation time to zero: ggilani 28/10/2014
 			Hosts[k].infector = -1;
@@ -2506,7 +2506,7 @@ void InitModel(int run) // passing run number so we can save run number in the i
 				Hosts[k].RecoveringFromCritical_time = USHRT_MAX - 1;
 				Hosts[k].Severity_Current = Severity::Asymptomatic;
 				Hosts[k].Severity_Final = Severity::Asymptomatic;
-				Hosts[k].inf = InfStat_Susceptible;
+				Hosts[k].inf = InfStat::Susceptible;
 			}
 		}
 
@@ -2724,7 +2724,7 @@ void SeedInfection(double t, int* NumSeedingInfections_byLocation, int rf, int r
 			for (k = 0; (k < NumSeedingInfections_byLocation[i]) && (m < 10000); k++)
 			{
 				l = Mcells[j].members[(int)(ranf() * ((double)Mcells[j].n))]; //// randomly choose member of microcell j. Name this member l
-				if (Hosts[l].inf == InfStat_Susceptible) //// If Host l is uninfected.
+				if (Hosts[l].inf == InfStat::Susceptible) //// If Host l is uninfected.
 				{
 					if (CalcPersonSusc(l, 0, 0, 0) > 0)
 					{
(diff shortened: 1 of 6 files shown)


zdroid:
rebased

src/Sweep.cpp
```diff
@@ -1330,7 +1330,7 @@ void DigitalContactTracingSweep(double t)
 						if ((Hosts[contact].dct_test_time == ts) && (Hosts[contact].index_case_dct == 0))
 						{
 							//if host is positive
-							if ((abs(Hosts[contact].inf) == 2) || (Hosts[contact].inf == -1)) //either asymptomatic infectious, symptomatic infectious or presymptomatic infectious
+							if ((abs((int) Hosts[contact].inf) == 2) || (Hosts[contact].inf == InfStat::InfectiousAlmostSymptomatic)) //either asymptomatic infectious, symptomatic infectious or presymptomatic infectious
```
mstevens-uk, line 1333: Replace the "2" with InfStat::InfectiousAsymptomaticNotCase

src/Sweep.cpp
```diff
@@ -1367,15 +1367,15 @@ void DigitalContactTracingSweep(double t)
 					else if (P.FindContactsOfDCTContacts)
 					{
 						//check every day to see if contacts become index cases - but they have to be infectious. Otherwise we could set the trigger time and cause their contacts to be traced when they are not being traced themselves.
-						if ((Hosts[contact].index_case_dct == 0) && ((abs(Hosts[contact].inf) == 2) || (Hosts[contact].inf == -1)))
+						if ((Hosts[contact].index_case_dct == 0) && ((abs((int) Hosts[contact].inf) == 2) || (Hosts[contact].inf == InfStat::InfectiousAlmostSymptomatic)))
```
mstevens-uk, line 1370: Replace the "2" with InfStat::InfectiousAsymptomaticNotCase

zdroid: But `Case` also gives abs value 2, as it was intended previously. Are you sure it's supposed to be replaced with that?

src/Sweep.cpp
weshinsley: I wonder about expanding the abs(...) == 2, to something explicitly stating the two cases - (InfStat_InfectiousAsymptomaticNotCase || InfStat_Case) - I think that would be clearer, and probably no worse than performing abs()

src/Sweep.cpp
weshinsley: Syntactially correct - abs() would result in InfStat::InfectiousAsymptomaticNotCase - but again see earlier comment - it may be clearer to expand these and remove the necessity for abs

src/CovidSim.cpp
```diff
@@ -2925,7 +2925,7 @@ int RunModel(int run, std::string const& snapshot_save_file, std::string const&
 								do
 								{
 									l = (int)(((double)P.PopSize) * ranf()); //// choose person l randomly from entire population. (but change l if while condition not satisfied?)
-								} while ((abs(Hosts[l].inf) == InfStat_Dead) || (ranf() > P.FalsePositiveAgeRate[HOST_AGE_GROUP(l)]));
+								} while ((abs((int) Hosts[l].inf) == (int) InfStat::Dead) || (ranf() > P.FalsePositiveAgeRate[HOST_AGE_GROUP(l)]));
 								DoFalseCase(l, t, ts, 0);
```
weshinsley, line 2929: Possibly Hosts[l].inf == one of {InfStat::Dead_WasAsymp, InfStat::Dead_WasSymp, InfStat::Dead}

mstevens-uk: OK, happy to withdraw the change request, as long as the abs code is addressed in the next PR. (unless you want to address it in this PR.)

weshinsley: I'd replace abs in this PR. It's going to touch a lot of the same lines.

weshinsley:
Also see PR #292 - which is replacing some of these comparisons with isDead() and isAlive()

zdroid:
@weshinsley @mstevens-uk I'd rather postpone further changes for the next PR, unless there is a clear way to avoid `abs`.

Ideally, the integer values and placeholders like `Dead` and `Recovered` should be removed altogether from InfStat, and replaced with a more idiomatic functionality.

PR #417, opened 2020-06-23
Refactor DigitalContactTracingSweep

DavidVernest:
Refactor the DigitalContactTracingSweep function in Sweep.cpp.

Specifically, extract four functions:

1. static void GetDctStartEndTimes(int infector, int contact, unsigned short contact_time, unsigned short& dct_start_time,
	unsigned short& dct_end_time);

2. void AddToDctQueue(int tn, int ad, int i3, int ci, unsigned short ts);

3. static void RemoveFromDctQueue(int j, int i, int k, int contact, int infector, unsigned short contact_time);

4. static void RemoveContactFromDctQueue(int contact);

### DIFF
--- a/src/Sweep.cpp
+++ b/src/Sweep.cpp
@@ -15,6 +15,10 @@
 #include "Sweep.h"
 #include "Update.h"
 
+void UpdateQueue(int j, int i, int k, int contact, int infector, unsigned short contact_time);
+void EndDigital(int contact);
+unsigned short GetStartTime(int infector, int contact, unsigned short contact_time);
+unsigned short GetEndTime(int infector, unsigned short& dct_start_time, int contact, unsigned short contact_time);
 
 void TravelReturnSweep(double t)
 {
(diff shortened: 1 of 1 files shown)


zdroid:
Maybe this should be moved out of Sweep.cpp? The file is quite large, so it may be useful to break it down.

mstevens-uk:
@zdroid Yes - plans are afoot to break both Sweep.cpp and Update.cpp into separate files/classes. But not in this PR - to keep things simpler for the reviewers, I have not broken up these files.

mstevens-uk:
@weshinsley @dlaydon @matt-gretton-dann Is Digital Contact Tracing actively used? If so, do we have any regression tests for it? It might not be worth refactoring otherwise - and I'll move on to something else. Thanks.

mstevens-uk:
@weshinsley  Hi - can I just confirm if Digital Contact Tracing is actively used? I can unit test and simplify the code - just looking to see if it is active before asking for a merge request.

dlaydon:
Hi @mstevens-uk, thanks for asking. We are using this code right now, but it isn't added to the formal regression tests. Therefore if you could refrain from changing or refactoring it for now that would be helpful. Thanks.

mstevens-uk:
Understood.

PR #418, opened 2020-06-23
Param vector updates

CloneDeath:
Converted a couple more fields in Param to BoundingBox.
Converted location of infected to a Vector2<double>.
Simple usage updates (will use more Vector math in a future PR, wanted to keep this one simple).

### DIFF
--- a/src/Geometry/Vector2.cpp
+++ b/src/Geometry/Vector2.cpp
@@ -2,6 +2,12 @@
 
 using namespace Geometry;
 
+Vector2<double> Geometry::operator+(const Vector2<double> &left, const Vector2<float> &right){
+	return left + Vector2<double>(right);
+}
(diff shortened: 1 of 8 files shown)


src/Geometry/Vector2.cpp
```diff
@@ -2,6 +2,12 @@
 
 using namespace Geometry;
 
+Vector2<double> Geometry::operator+(const Vector2<double> &left, const Vector2<float> &right){
+	return left + Vector2<double>(right);
+}
+Vector2<double> Geometry::operator+(const Vector2<float> &left, const Vector2<double> &right){
+	return Vector2<double>(left) + right;
```
zdroid, line 9:
```suggestion
	return right + Vector2<double>(left);
```
Uses associativity of addition here. In that case, if you change something in one of the two, you don't have to change the other one.

PR #421, opened 2020-06-23
Use better if-else-elseif flow for model CalibIteration

zdroid:
It's all only refactoring, except this one:
```cpp
if ((P.ModelCalibIteration >= 2) && ((((RatioPredictedObserved - 1.0) <= DesiredAccuracy) && (RatioPredictedObserved >= 1)) || (((1.0 - RatioPredictedObserved) <= DesiredAccuracy) && (RatioPredictedObserved < 1))))
  P.StopCalibration = 1;
```

Based on my understanding, what it is trying to achieve is equivalent to:
```cpp
if (abs((RatioPredictedObserved - 1.0) <= DesiredAccuracy)) {
  P.StopCalibration = 1;
}
```
Edit: tests failed with that change, so I reverted it.

### DIFF
--- a/src/CovidSim.cpp
+++ b/src/CovidSim.cpp
@@ -5104,7 +5104,8 @@ void RecordSample(double t, int n)
 						if (k > 0) P.CaseOrDeathThresholdBeforeAlert = k;
 					}
 					else if (P.ModelCalibIteration >= 2) {
-						if (abs((RatioPredictedObserved - 1.0) <= DesiredAccuracy)) {
+						if ((((RatioPredictedObserved - 1.0) <= DesiredAccuracy) && (RatioPredictedObserved >= 1)) ||
+							(((1.0 - RatioPredictedObserved) <= DesiredAccuracy) && (RatioPredictedObserved < 1))) {
 							P.StopCalibration = 1;
 						}
 						else if (P.ModelCalibIteration % 2 == 0) // on even iterations, adjust timings
--- a/src/CovidSim.cpp
+++ b/src/CovidSim.cpp
@@ -5100,7 +5100,7 @@ void RecordSample(double t, int n)
 					if (P.ModelCalibIteration == 0)
 					{
 						// rescale threshold
-						k = (int)(((double)P.CaseOrDeathThresholdBeforeAlert) / RatioPredictedObserved);
+						k = (int)(P.CaseOrDeathThresholdBeforeAlert / RatioPredictedObserved);
 						if (k > 0) P.CaseOrDeathThresholdBeforeAlert = k;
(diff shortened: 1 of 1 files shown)


src/CovidSim.cpp
```diff
@@ -5097,29 +5097,34 @@ void RecordSample(double t, int n)
 					if (DesiredAccuracy < 0.05) DesiredAccuracy = 0.05;
 					fprintf(stderr, "\n** %i %lf %lf | %lg / %lg \t", P.ModelCalibIteration, t, P.DateTriggerReached_SimTime + P.DateTriggerReached_CalTime - P.Interventions_StartDate_CalTime, P.HolidaysStartDay_SimTime,RatioPredictedObserved);
 					fprintf(stderr, "| %i %i %i %i -> ", trigAlert, trigAlertCases, P.AlertTriggerAfterIntervThreshold, P.CaseOrDeathThresholdBeforeAlert);
-					if ((P.ModelCalibIteration >= 2) && ((((RatioPredictedObserved - 1.0) <= DesiredAccuracy) && (RatioPredictedObserved >= 1)) || (((1.0 - RatioPredictedObserved) <= DesiredAccuracy) && (RatioPredictedObserved < 1))))
-						P.StopCalibration = 1;
-					else if (P.ModelCalibIteration == 0)
+					if (P.ModelCalibIteration == 0)
 					{
 						// rescale threshold
 						k = (int)(((double)P.CaseOrDeathThresholdBeforeAlert) / RatioPredictedObserved);
```
weshinsley, line 5103: RatioPredictedObserved is double, so this can just be
k = (int) (P.CaseOrDeathThresholdBeforeAlert / RatioPredictedObserved);

src/CovidSim.cpp
```diff
@@ -5097,29 +5097,34 @@ void RecordSample(double t, int n)
...
+								P.Epidemic_StartDate_CalTime--;
+								P.HolidaysStartDay_SimTime--;
+							}
+							else if (RatioPredictedObserved < 1) // ... otherwise if too few cases/deaths,  make timings later
+							{
+								P.Epidemic_StartDate_CalTime++;
+								P.HolidaysStartDay_SimTime++;
+							}
+						}
+						else { // on odd iterations, adjust seeding.
```
weshinsley, line 5124: Perhaps write this as something like
```
else /* if (P.ModelCalibIteration % 2 == 1) */ 
{
```

zdroid: Yeah that's alright.

weshinsley:
This has been part-implemented, and part-changed by Neil's recent work on PR #430

PR #428, opened 2020-06-25
Allow Clang 9 To Build Master

ozmorph:
#404 introduced 4x compiler errors that prevents Clang 9 from building master.

```bash
ozmorph@home:~/covid-sim/build$ clang++-9 --version
clang version 9.0.1-12 
Target: x86_64-pc-linux-gnu
Thread model: posix
```

```bash
/home/ozmorph/covid-sim/src/Update.cpp:1356:33: error: ordered comparison between pointer and zero ('int *' and 'int')
        assert(Cells[cellIndex].latent >= 0);
               ~~~~~~~~~~~~~~~~~~~~~~~ ^  ~
/usr/include/assert.h:93:27: note: expanded from macro 'assert'
     (static_cast <bool> (expr)                                         \
                          ^~~~
/home/ozmorph/covid-sim/src/Update.cpp:1357:35: error: ordered comparison between pointer and zero ('int *' and 'int')
        assert(Cells[cellIndex].infected >= 0);
               ~~~~~~~~~~~~~~~~~~~~~~~~~ ^  ~
/usr/include/assert.h:93:27: note: expanded from macro 'assert'
     (static_cast <bool> (expr)                                         \
                          ^~~~
/home/ozmorph/covid-sim/src/Update.cpp:1368:33: error: ordered comparison between pointer and zero ('int *' and 'int')
        assert(Cells[cellIndex].latent >= 0);
               ~~~~~~~~~~~~~~~~~~~~~~~ ^  ~
/usr/include/assert.h:93:27: note: expanded from macro 'assert'
     (static_cast <bool> (expr)                                         \
                          ^~~~
/home/ozmorph/covid-sim/src/Update.cpp:1380:35: error: ordered comparison between pointer and zero ('int *' and 'int')
        assert(Cells[cellIndex].infected >= 0);
               ~~~~~~~~~~~~~~~~~~~~~~~~~ ^  ~
/usr/include/assert.h:93:27: note: expanded from macro 'assert'
     (static_cast <bool> (expr)                                         \
                          ^~~~
2 warnings and 4 errors generated.
```

This patch fixes it.

### DIFF
--- a/src/Update.cpp
+++ b/src/Update.cpp
@@ -1353,8 +1353,6 @@ static void SusceptibleToRecovered(int cellIndex)
 
 	// assert values are non-negative
 	assert(Cells[cellIndex].S >= 0);
-	assert(Cells[cellIndex].latent >= 0);
-	assert(Cells[cellIndex].infected >= 0);
 }
 
 static void SusceptibleToLatent(int cellIndex)
@@ -1365,7 +1363,6 @@ static void SusceptibleToLatent(int cellIndex)
 
 	// assert values are non-negative
 	assert(Cells[cellIndex].S >= 0);
-	assert(Cells[cellIndex].latent >= 0);
 }
 
 
@@ -1377,7 +1374,6 @@ static void LatentToInfectious(int cellIndex)
 
 	// assert values are non-negative
 	assert(Cells[cellIndex].L >= 0);
-	assert(Cells[cellIndex].infected >= 0);
 
 }
 


src/Update.cpp
```diff
@@ -1353,8 +1353,8 @@ static void SusceptibleToRecovered(int cellIndex)
 
 	// assert values are non-negative
 	assert(Cells[cellIndex].S >= 0);
-	assert(Cells[cellIndex].latent >= 0);
-	assert(Cells[cellIndex].infected >= 0);
+	assert(Cells[cellIndex].latent >= static_cast<int*>(0));
+	assert(Cells[cellIndex].infected >= static_cast<int*>(0));
```
mstevens-uk, line 1357: Should we remove the asserts against int* pointers? They were only added as an extra while asserting the severity states, which are all integers.

mstevens-uk, line 1357: Thanks, but this doesn't completely fix the issue. 
If the pointer is decremented too far, the memory address is still positive. 
It would be safer to remove these asserts against the int* pointers - they can be re-added in a future PR if necessary.

DavidVernest (changes requested):

src/Update.cpp
ozmorph: Makes sense. Removed as requested. Thanks for the review.

matt-gretton-dann: Removing is fine for this PR

I think ultimately the right check is to test that `Cells[cellIndex].infected` points somewhere within the range [`Cells[cellindex].susceptible`, Cells[cellindex].susceptible + Cells[cellindex].n`).

matt-gretton-dann (approved):
Approved.

PR #433, opened 2020-07-02
Regression Test Heap Optimization: Remove Extinct Result Arrays

ozmorph:
**Objective**

Reduce RAM usage in regression tests so that GitHub Action runners can run all 4 tests without paging memory.

**Background**

The size of the `Results` struct in `Model.h` is nearly 805 KB on Clang 9 for Linux, which when accounting for an array of `P.NumSamples` (`720` in the regression tests) means that `TSMeanE` and `TSVarE` each require an allocation of around 580 MB. As a result, a single regression test will use over 1.1 GB of RAM just for these two arrays alone. When all 4 tests are ran in parallel, they consume 4.4 GB of RAM.

**Rationale**

As noted in #388, the default GitHub action runners have only 7 GB of total RAM and a 14 GB temporary storage device allocated to them. For Linux environments [the swap space is set to 4GB](https://github.com/actions/virtual-environments/pull/965) and on Windows [the swap space is most likely set to 2GB](https://github.com/mrc-ide/covid-sim/pull/388#issuecomment-646626562).

Local runs using [heaptrack](https://github.com/KDE/heaptrack) indicate that about 3.5GB is the average heap size of each run of `CovidSim`. If accounting for the fact that the US regression tests run for a fraction of the time that the UK tests do, let's assume that on average 7 GB of RAM is being used by the regression tests and a maximum of 14 GB is used. Windows Server 2019 requires [512MB of RAM to run](https://docs.microsoft.com/en-us/windows-server/get-started-19/sys-reqs-19). If 7 GB is the total amount of RAM given to the runners, this does not leave much room to the operating system itself. Therefore, I surmise that `CovidSim` is hitting paged memory for at least the first half of a `ctest -j6 -V` run.

**Solution**

This pull request removes the `TSMeanE` and `TSVarE` arrays from the code since they haven't been [used in the final output](https://github.com/mrc-ide/covid-sim/blame/8c4f0d547b5fe13cb300adcb2555fde9fc203dc4/src/CovidSim.cpp#L365) since the initial squash of commits back in April. The regression tests still pass.

**Known Concerns**

If this code is still used on the super-computers or outside of the regression tests listed on GitHub, then please close this PR and I will open a new PR that documents its use so that someone doesn't attempt to propose this change in the future.

**Alternative/Additional Solution**

The memory requirements of the `Result` arrays could also be optimized either with heap-allocated C style arrays or std::vectors  instead of static double arrays of `MAX_ADUNITS` size (which is set to `3200` but the regression tests only use 1 or 2 elements). However this would require far deeper and more extensive changes to the code base than this Pull Request.

### DIFF
--- a/src/CovidSim.cpp
+++ b/src/CovidSim.cpp
@@ -89,8 +89,8 @@ Place** Places;
 AdminUnit AdUnits[MAX_ADUNITS];
 //// Time Series defs:
 //// TimeSeries is an array of type results, used to store (unsurprisingly) a time series of every quantity in results. Mostly used in RecordSample.
-//// TSMeanNE and TSVarNE are the mean and variance of non-extinct time series. TSMeanE and TSVarE are the mean and variance of extinct time series. TSMean and TSVar are pointers that point to either extinct or non-extinct.
-Results* TimeSeries, * TSMean, * TSVar, * TSMeanNE, * TSVarNE, * TSMeanE, * TSVarE; //// TimeSeries used in RecordSample, RecordInfTypes, SaveResults. TSMean and TSVar
+//// TSMean and TSVar are the mean and variance of non-extinct time series.
+Results* TimeSeries, * TSMean, * TSVar; //// TimeSeries used in RecordSample, RecordInfTypes, SaveResults. TSMean and TSVar
 Airport* Airports;
 BitmapHeader* bmh;
 //added declaration of pointer to events log: ggilani - 10/10/2014
@@ -351,15 +351,11 @@ int main(int argc, char* argv[])
 	}
 
 	P.NRactual = P.NRactNE;
-	TSMean = TSMeanNE; TSVar = TSVarNE;
 	if ((P.DoRecordInfEvents) && (P.RecordInfEventsPerRun == 0))
 	{
 		SaveEvents(output_file_base);
 	}
 	SaveSummaryResults(output_file_base + ".avNE");
-	P.NRactual = P.NRactE;
-	TSMean = TSMeanE; TSVar = TSVarE;
-	//SaveSummaryResults(output_file_base + ".avE");
 
 	Bitmap_Finalise();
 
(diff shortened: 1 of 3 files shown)


ozmorph:
Sounds good! I like having a check as an alternative.

I am surprised too that it did not affect run-time. It could also be that the paging algorithm and Windows allocators are good enough to hide this cost. I'm going to try to replicate the CI binaries on my Windows system and analyze the performance versus my Linux runs to try to determine where the real difference lies.

Good to note about running only 2 tests at once on the CI.

ozmorph:
@matt-gretton-dann Using the exact same compiler that GitHub actions is using, I get similar performance on my local machine as Clang 9 on WSL Ubuntu 20.04 does. This led me down a rabbit hole, but I think your past theory about all of the `fprintf()` calls being responsible for the slowdown are probably correct.

It appears that I/O performance is a common issue for GitHub actions (especially on Windows):

- [Database operations were much slower](https://github.community/t/insight-into-vm-performance/16199)
- [Windows is 3.5x slower than Linux](https://github.community/t/performance-on-windows/16984)
- [Windows I/O container is slower than Linux container](https://github.community/t/windows-container-i-o-performance-is-slower-than-on-a-linux-container/16897)

It appears this project is not the only one to be having this issue.  It also explains why Windows performance is so drastically different than Linux and Mac.

To conclude, reducing memory usage anywhere is not going to make the regression tests faster by itself but I think it would provide the extra headroom necessary for more variable tracking in additional simulations in the future.

matt-gretton-dann:
> This led me down a rabbit hole, but I think your past theory about all of the fprintf() calls being responsible for the slowdown are probably correct.

I have a [WiP branch](https://github.com/mrc-ide/covid-sim/tree/matt-gretton-dann/rework-printing) which enables us to control output verbosity - and so that should show whether I am right or not.

Issue #424, opened 2020-06-24
You can't multiply two vectors to get another vector


zebmason:
https://github.com/mrc-ide/covid-sim/blob/f6d40d605feeaac0d18ae256875808df42eecf92/src/Geometry/Vector2.cpp#L5

This has no basis in vector algebra so is dangerous

insidedctm:
The operation results in both components of the Vector2 being squared. I'd guess this is a precursor to doing something like a distance calculation. How is it used in the main code? This should either be renamed or maybe a change to the interface to make it more obvious what the Vector2 operation is actually being used to do.

zebmason:
The Vector2 class is used to represent both a vector and a diagonal matrix hence the mixed algebra

PR #436, opened 2020-07-09
Set scale to be a diagonal matrix and not a vector.

zebmason:
Change the vector multiplication and division operations to use a diagonal matrix.

This fixes issue #424


zebmason:
@weshinsley @dlaydon  This fixes the Mathematics error vector times vector equals vector - as a Mathematician I find such errors offensive.

PR #437, opened 2020-07-09
Delete the example test now that there are some tests that do something

zebmason:
Tidy up


zebmason:
Remove example code introduced in #324 now that #408 has been merged with its tests.

PR #440, opened 2020-07-19
Split the unit and regression tests

zebmason:


zebmason:
@matt-gretton-dann You should be interested in my changes to make the container matrix three dimensional. When I did the CMake upgrade I should have made it two dimensional. Obviously it means that seven of the "Required" tests are no longer run as their names have changed.

zebmason:
Changed from 
```    strategy:
      fail-fast: false
      matrix:
        container_image: ["fedora:latest", "debian:10", "archlinux/base", "opensuse/tumbleweed", "alpine:latest"]
        cmake: ["cmake"]
        ctest: ["ctest"]
        include:
        - container_image: "ubuntu:bionic"
          cmake: "../cmake-3.17.2-Linux-x86_64/bin/cmake"
          ctest: "ctest"
        - container_image: "centos:7"
          cmake: "cmake"
          ctest: "ctest3"
```
as that changes the test names (all 3 parts of the matrix are included in the name) leading GitHub hanging for the original test names to complete.

weshinsley (approved):

PR #441, opened 2020-07-24
Separate interface from implementation as per the normal convention

zebmason:
Further merging of #213


zebmason:
Obviously if this is not a convention that you want to follow I'll withdraw this.

covid-sim.vcxproj
```diff
@@ -130,9 +130,9 @@
     <ClInclude Include="src\Update.h" />
     <ClInclude Include="src\MicroCellPosition.hpp" />
     <ClInclude Include="src\Direction.hpp" />
-    <ClInclude Include="src\Geometry\BoundingBox.h" />
-    <ClInclude Include="src\Geometry\Size.h" />
-    <ClInclude Include="src\Geometry\Vector2.h" />
+    <ClInclude Include="include\Geometry\BoundingBox.h" />
+    <ClInclude Include="include\Geometry\Size.h" />
+    <ClInclude Include="include\Geometry\Vector2.h" />
     <ClInclude Include="src\Models\Cell.h" />
```
weshinsley, line 136: Should these have lower-case G on geometry? (Not that Windows really cares...)

zebmason: Good point, pays to be consistent. I'll change.

PR #444, opened 2020-08-02
Separate out the distance calculations

zebmason:
By removing branching this may make the calculations faster (branch prediction may disagree). Can add in unit tests.

@weshinsley Let me know if you are interested and I'll do further development.


zebmason:
This fixes #447

PR #445, opened 2020-08-03
InfStat and Direction rollup of refactors

weshinsley:
This is a consolidation/completion of a number of PRs, mainly dealing with InfStat, making Host->inf private, resolving the issues where we had numerical comparisons or abs(x) for enum states. At the end of the refactor, the Dead and Recovered states are no longer referenced - only their symptomatic / asymptomatic classifications are.

So, I think this PR brings together ideas from dead PR #416 and PR #292 - which was approved at one time, but further work on it got a bit overwhelming I think, and it is quite stale now.

This PR also includes changes from #415 and #439, (both dead) as I had those ready to go, but missed the boat over Summer.

It resolves issue #291 and the typo emphatically reported in #456

And it comes with documentation and unit tests, so while the code is not tremendously exciting, hopefully it demonstrates useful practice on the way.


weshinsley:
Taking this back to draft while I try and write some (very simple) tests...

NeilFerguson (approved):

weshinsley:
The previous merge has just conflicted a couple of lines and I've not got the resolution quite right - just fixing...

Issue #452, opened 2020-09-02
Fail early on regression tests?


zebmason:
The unit and regression tests have just been split from one another. If I remember rightly this means that ctest is always used to run the tests and that it is possible to get the tests to stop on the first failed test.

If there is any interest I'll get the regression tests to stop on the first failed test - the unit tests will all run pass or fail.

weshinsley:
Yes, I think that would be good - I was wondering if that might be possible, since unit tests are likely to fail (or indeed, succeed) faster than the regression tests...

PR #453, opened 2020-09-03
Stop on first regression test failure

zebmason:
Address #452


zebmason:
@weshinsley Makes a difference on Windows and Mac as they use version 3.18 of CTest. On Linux CTest is older and doesn't support the option but doesn't seem to object to it appearing on the command line.

zebmason:
Updated CTest on Linux now as in same archive as CMake

zebmason:
Couldn't get alpine:latest to fail early

weshinsley (approved):

PR #454, opened 2020-09-28
Resolve many VS warnings including mystery 'k'

weshinsley:
Not a very exciting PR, but this deals with all the [C26451](https://docs.microsoft.com/en-us/cpp/code-quality/c26451?view=vs-2019) warnings VS gives about potential arithmetic overflows. (I don't think we have any, but this at least clears out the noise).

More excitingly (to me), it deals with the 
`1>C:\Files\Dev\covid-sim\src\SetupModel.cpp(770): warning C4101: 'k': unreferenced local variable`

where there is clearly no reference to k on line 770
```
void SetupPopulation(std::string const& density_file, std::string const& out_density_file, std::string const& school_file, std::string const& reg_demog_file)
```

The warning actually refers to an OpenMP section where `k` is not needed and should be removed from line 1283

```
#pragma omp parallel for private(j,k,m,s) schedule(static,1) default(none) \
```

### DIFF
--- a/src/Constants.h
+++ b/src/Constants.h
@@ -100,4 +100,9 @@ const int HOTELS_PER_1000PASSENGER = 10;
 
 const int MAX_NUM_INTERVENTION_CHANGE_TIMES = 100; // may want to make this intervention-specifc, but keep simple for now.
 
+// MS generates a lot of C26451 overflow warnings. Below is shorthand
+// to increase the cast size and clean them up.
+
+#define _I64(x) static_cast<int64_t>(x)
+
 #endif // COVIDSIM_CONSTANTS_H_INCLUDED_
--- a/src/CovidSim.cpp
+++ b/src/CovidSim.cpp
@@ -460,11 +460,11 @@ void ReadParams(std::string const& ParamFile, std::string const& PreParamFile, s
 	if (P.FitIter == 0)
 	{
 		for (i = 0; i < MAX_COUNTRIES; i++) {
-			CountryNames[i] = CountryNameBuf + _I64(128) * i;
+			CountryNames[i] = CountryNameBuf + INT64_C(128) * i;
 			CountryNames[i][0] = 0;
 		}
 		for (i = 0; i < MAX_ADUNITS; i++) {
-			AdunitListNames[i] = AdunitListNamesBuf + _I64(128) * i;
+			AdunitListNames[i] = AdunitListNamesBuf + INT64_C(128) * i;
 			AdunitListNames[i][0] = 0;
 		}
 		for (i = 0; i < 100; i++) P.clP_copies[i] = 0;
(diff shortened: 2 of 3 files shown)


src/Constants.h
```diff
@@ -100,4 +100,9 @@ const int HOTELS_PER_1000PASSENGER = 10;
 
 const int MAX_NUM_INTERVENTION_CHANGE_TIMES = 100; // may want to make this intervention-specifc, but keep simple for now.
 
+// MS generates a lot of C26451 overflow warnings. Below is shorthand
+// to increase the cast size and clean them up.
+
+#define _I64(x) static_cast<int64_t>(x)
```
matt-gretton-dann, line 106: Where you are using this for constants I would instead do `#include <cstdint>` and then use `INT64_C()` and `UINT64_C()` as appropriate.

Where you are using this for variables I'd be interested to see why the original variable wasn't a 64-bit value to begin with.

weshinsley: Thanks - I've changed constants to use INT64_C - all involved signed ints.

So, if I understand the VS warning correctly... 

`base = (IndexList*)Memory::xcalloc(P.NMCP * NNA, sizeof(IndexList));`
where P.NCMP and NNA are both ints, and xcalloc takes (size_t, size_t) - size_t being 64-bits on a 64-bit machine - (I think?) - the warning message is...

Warning C26451: Arithmetic overflow: Using operator '*' on a 4 byte value and then casting the result to a 8 byte vlaue. Cast the value to the wider type before calling operator '*' to avoid overflow (io.2).

So xcalloc can take an 8-byte size; we know P.NMCP * NNA is never going to need 8 bytes here, but VS is saying that if P.NMCP * MNA needed 8-bytes, we would overflow in the maths before we got the 8-byte value to call xcalloc with. (We know only from prior knowledge that we'll not need 8-byte here). I suspect using an alternative xcalloc that takes 4-byte types would be another way to appease VS here, as then there would be no point in upscaling, and VS will just leave us alone to decide for ourselves whether the 4-byte mult could overflow.

It also warns on... 
division: eg, double / (int - int)  - it prefers (double) / (int64 - int)
cast: eg, (double) (int - int) - it prefers (double) (int64 - int)

or similarly, calling ignpoi_mt(int - int) - where ignpoi_mt expects a double. To make that warning-free, I needed to a slightly awkward ignpoi_mt((double) (_I64(int) - int)) - if I remove the extra (double) cast, then I instead get a C4244 - conversion from int64_t to double is a possible loss of data.

I'm not sure of best approach here - the VS warnings are not especially helpful for any of these arith cases, and my fixes have been about "making the warnings go away" rather than anything more intelligent. At the same time, warning-free clean compilation remains a noble aim...

Keen to know your thoughts/whether this is worth it, or whether there's a better way by code or by sensible warning suppression... (I like the 'k' fix though - would be one of my best 2-byte PRs!)

matt-gretton-dann: > So xcalloc can take an 8-byte size_t; we know P.NMCP * NNA is never going to need 8 bytes here, but VS is saying that if P.NMCP * MNA needed 8-bytes, we would overflow in the maths before we got the 8-byte value to call xcalloc with.

Does putting an assert in just before help the compiler?  Something like:
```c++
assert(P.NMCP < (INT64_MAX / MNA)); /* Silence compiler warning. */
```
This states the knowledge we have, and should also alert us if we do break that rule.

> It also warns on...
> division: eg, double / (int - int) - it prefers (double) / (int64 - int)
> cast: eg, (double) (int - int) - it prefers (double) (int64 - int)

This does start looking like an over-active compiler. The other alternative to _I64() is to again see if asserts() help to prove we won't overflow.

But I'm happy with what you have.

matt-gretton-dann (approved):

src/Constants.h
weshinsley: The assert unfortunately doesn't prevent the warning... boo....
Yes, I can see the point, but over-zealous compiler warnings I think...

PR #465, opened 2021-02-05
Initialize ModelValue properly in CovidSim.cpp

krook1024:
The compiler probably handles it, but let us be on the safe side.


e4lam:
> The compiler probably handles it, but let us be on the safe side.

No, absolutely not. The compiler will **NOT** handle this. C++ does not work that way. You might be lucky that it's usually zero just because of the way OSes work but it make no mistake: this initialization is needed for correctness.

weshinsley:
Thank you - good spot

weshinsley (approved):

PR #469, opened 2021-02-17
Dlaydon/mcmc 2

dlaydon:


weshinsley (approved):
Looks good - Perhaps check variable naming convention re: StepDown vs Stepdown on variables later - but let's get things moving again for now.

PR #480, opened 2021-03-18
IFR scales linearly not step function

dlaydon:


weshinsley:
Implemented on another branch

PR #491, opened 2022-05-26
Vary contact rates2

dlaydon:
Consolidated multiple NPI parameters into single large P.Efficacies array. So far does nothing except reproduce existing functionality. Idea is that it will form basis from which to vary interventions over time and across regions.

### DIFF
--- a/src/Constants.h
+++ b/src/Constants.h
@@ -1,6 +1,8 @@
 #ifndef COVIDSIM_CONSTANTS_H_INCLUDED_
 #define COVIDSIM_CONSTANTS_H_INCLUDED_
 
+#include "Country.h"
+
 /**
  * Math constant defined as the ratio of a circle's circumference to its diameter.
  *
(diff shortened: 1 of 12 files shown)


src/CovidSim.cpp
```diff
@@ -293,6 +290,12 @@ int main(int argc, char* argv[])
 
 	Files::xfprintf_stderr("Model setup in %lf seconds\n", ((double) clock() - cl) / CLOCKS_PER_SEC);
 
+	// Allocate memory for Efficacies array
+	P.NumInfectionSettings		= P.NumPlaceTypes + 2;	// Household, Place (x P.NumPlaceTypes), Spatial / Community
```
weshinsley, line 294: Make this NUM_PLACE_TYPES + 2

src/Constants.h
```diff
@@ -107,4 +107,21 @@ const int MAX_NUM_CFR_CHANGE_TIMES = 100; /**< To allow IFR to scale over time.
 
 #define _I64(x) static_cast<int64_t>(x)
 
+// Settings (numbers not arbitrary - don't change without checking)
+const int PrimarySchool		= 0; 
+const int SecondarySchool	= 1; 
+const int University		= 2; 
+const int Workplace			= 3; 
+const int House				= 4;
```
weshinsley, line 115: const int House = NUM_PLACE_TYPES;

const int Spatial = NUM_PLACE_TYPES + 1;

weshinsley, line 115: const int House = NUM_PLACE_TYPES;

const int Spatial = NUM_PLACE_TYPES + 1;

dlaydon: Don't think this will work, would work if const int House = P.NumPlaceTypes and const int Spatial = P.NumPlaceTypes +1; but don't think that will work with const

weshinsley: The above will work, but will waste a tiny bit of memory if P.NumPlaceTypes turns out less than NUM_PLACE_TYPES. Alternative is to declare it as non-const, and then set it once you know what P.NumPlaceTypes is - but I'm not sure that is worth the bother in this case.

weshinsley (approved):
Fixed array out of bounds, tests all pass. 

Rest looks fine - could think further about best way of handling different numbers of place-types, and model-specific meanings of them.

